PR #1, opened 2023-09-22
setup

zklgame:
This MR did:
1. setup the service and enable the first POST request `http://localhost:8801/api/v1/process/start`

Compare to [iwf](https://github.com/indeedeng/iwf), reduce the abstraction level for better readability. E.g., combine the API service inteface, service, and handler together.


service/api/handler.go
```diff
@@ -0,0 +1,54 @@
...
+
+func newHandler(config config.Config, logger log.Logger) *handler {
+	return &handler{
+		config: config,
+		logger: logger,
+	}
+}
+
+func (h *handler) ApiV1ProcessStartPost(c *gin.Context) {
+	var req xdbapis.ObjectExecutionStartRequest
```
zklgame, line 26: Might rename the request to be `ProcessExecutionStartRequest` ?

longquanzheng (approved):

service/api/handler.go
longquanzheng: Yeah we need to fix the api repo. 

Also, I realized that moving the gen file in the api repo so that this server repo and Golang SDK repo will look cleaner. I saw some other projects done that recently

PR #30, opened 2023-10-01
Implement basic features for state execution

longquanzheng:


service/api/interfaces.go
```diff
@@ -5,6 +5,12 @@ import (
...
 
+type Server interface {
+	// Start will start running on the background
+	Start() error
+	Stop(ctx context.Context) error
+}
+
 // Service is the interface of API service, which decoupled from REST server framework like Gin
 // So that users can choose to use other REST frameworks to serve requests
 type Service interface {
```
zklgame, line 16: Can we reduce the the abstraction of interface here given these 2 reasons:
1. we only have one service, and not likely to suppor one more service with the same interface.
2. the goland ide could not directly jump to the implementation with this interface layer. Removing it will make it easier to search the codes in IDE.

extensions/postgres/error_checker.go
```diff
@@ -4,20 +4,22 @@ import (
...
 	"errors"
+	"fmt"
 	"github.com/lib/pq"
 )
 
-// ErrDupEntry indicates a duplicate primary key i.e. the row already exists,
+// ErrDupEntryCode indicates a duplicate primary key i.e. the row already exists,
 // check http://www.postgresql.org/docs/9.3/static/errcodes-appendix.html
-const ErrDupEntry = "23505"
+const ErrDupEntryCode = pq.ErrorCode("23505")
```
zklgame, line 13: IIUC, unique_violation will occur when attempting to insert a new row with values ("namespace", "process_id", "2") into the db, while the db already has a row with values ("namespace", "process_id", "1").

Then I think we need to modify the logic here later to support the idReusePolicy?

zklgame: Oh I see the comment.

service/api/interfaces.go
longquanzheng: > the goland ide could not directly jump to the implementation with this interface layer. Removing it will make it easier to search the codes in IDE.

It actually works for me. Lets sync up on this later. 

Unlike some lang like Java, adding interface to struct is quite standard in Golang. For example, in Java you can just write unit tests to mock a class. In Golang, you cannot mock any struct. You have to add an interface and mock the interface instead. 

Interfaces are needed not for unit tests, but also for cleaner code. It's important to have an interface if the code is being shared to other package to use. For example, the Server interface is being used in `bootstrap`. This `Service` interface is going to be used for Indeed internal implementation of Server -- because internally we use other built in REST framework which comes with metrics/logging.

persistence/sql/sql_process_store.go
```diff
@@ -0,0 +1,540 @@
...
+		err = tx.InsertAsyncStateExecution(ctx, stateRow)
+		if err != nil {
+			return nil, err
+		}
+
+		workerTaskRow := extensions.WorkerTaskRowForInsert{
+			ShardId:            request.NewTaskShardId,
+			ProcessExecutionId: prcExeId,
+			StateId:            req.GetStartStateId(),
+			StateIdSequence:    1,
```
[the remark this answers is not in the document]
longquanzheng: yeah it doesn't have to because compiler will know it its int32.

extensions/postgres/schema/xdb_sys_schema.sql
```diff
@@ -0,0 +1,61 @@
...
+    PRIMARY KEY (namespace, process_id)
+);
+
+CREATE TABLE xdb_sys_process_executions(
+    namespace VARCHAR(15) NOT NULL, -- for quick debugging
+    id uuid NOT NULL,
+    process_id VARCHAR(255) NOT NULL,
+    --
+    is_current BOOLEAN NOT NULL , -- for quick debugging
+    status SMALLINT, -- 0:undefined/1:running/2:completed/3:failed/4:timeout/5:terminated
```
duoertai, line 14: enum?

engine/worker_task_queue.go
```diff
@@ -0,0 +1,165 @@
...
+		processor:                 processor,
+		tasksToCommitChan:         make(chan persistence.WorkerTask, qCfg.ProcessorBufferSize),
+		currentReadCursor:         0,
+		pendingTaskSequenceToPage: make(map[int64]*workerTaskPage),
+	}
+}
+
+func (w *workerTaskQueueSQLImpl) Stop(ctx context.Context) error {
+	// a final attempt to commit the completed page
+	return w.commitCompletedPages(ctx)
```
zklgame, line 60: The processed task in incompleted pages will be executed again when starting the server next time?

duoertai (approved):

extensions/postgres/schema/xdb_sys_schema.sql
```diff
@@ -0,0 +1,61 @@
+CREATE TABLE xdb_sys_current_process_executions(
```
duoertai, line 1: what is the purpose of this table?

engine/worker_task_conrrent_processor.go
```diff
@@ -0,0 +1,295 @@
...
+func (w *workerTaskConcurrentProcessor) Start() error {
+	concurrency := w.cfg.AsyncService.WorkerTaskQueue.ProcessorConcurrency
+
+	for i := 0; i < concurrency; i++ {
+		go func() {
+			for {
+				select {
+				case <-w.rootCtx.Done():
+					return
+				case task, ok := <-w.taskToProcessChans:
```
zklgame, line 74: Question: IIUC, currently, the task in taskToProcessChans is executed sequentially, not in parallel?

zklgame (approved):

extensions/postgres/schema/xdb_sys_schema.sql
longquanzheng: yea technically we could use Postgres enum. 
However, it is difficult to write the code with enum if the row struct doesn't have a string type for this status.
If the row struct uses an enum, it will make it difficult for other database like SQLServer/Oracle to write code with it. 

This is already the case when I use UUID in Postgres for ProcessExecutionId. In MySQL, there is no UUID type(it has string with UUID helpers). So I have to make a ProcessExecutionIdString in addition to ProcessExecutionId in the struct. But UUID is quite important because otherwise using BYTEA of Postgres will make the debugging/query in psql much harder. So that additional field is worthy.

So to make it easier for all the database types, I would suggest not to use enum here.

engine/worker_task_queue.go
longquanzheng: Yes that's expected. When it executes again, it will load the AsyncStateExecution and found that it's already executed, then it will skip.

extensions/postgres/schema/xdb_sys_schema.sql
longquanzheng: It's for ProcessIdReusePolicy,  checking which executionId is the current/latest one for the same processId. 
https://github.com/xdblab/xdb/wiki/XDB-Proposal-&-Design#xdb_sys_process_current_executions


This is similar to Cadence/Temporal's IdReusePolicy.

engine/worker_task_conrrent_processor.go
longquanzheng: It will be in parallel because there are `concrrency` number of goroutines running to wait for the task. So the tasks will be distributed among all the goroutines and execute in paralllel

common/uuid/uuid.go
```diff
@@ -0,0 +1,111 @@
...
+	return string(buf[:])
+}
+
+// Scan implements sql.Scanner interface to allow this type to be
+// parsed transparently by database drivers
+func (u *UUID) Scan(src interface{}) error {
+	if src == nil {
+		return nil
+	}
+	guuid := &uuid.UUID{}
```
zhxnlai, line 87: I've heard some good things about UUIDv7. https://buildkite.com/blog/goodbye-integers-hello-uuids

Here is a popular implementation in Go: https://github.com/jetpack-io/typeid

common/uuid/uuid.go
```diff
@@ -0,0 +1,111 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package uuid
+
+import (
+	"database/sql/driver"
+	"encoding/hex"
+
+	"github.com/google/uuid"
```
zhxnlai, line 27: I've heard some good things about UUIDv7. https://buildkite.com/blog/goodbye-integers-hello-uuids

Here is a popular implementation in Go: https://github.com/jetpack-io/typeid

longquanzheng: Good to know. I will try it out later

PR #35, opened 2023-10-08
Refactor for insertAsyncStateExecution and insertWorkerTask functions

zklgame:
This MR did:
Extract common logic used in two places into shared functions and reduce duplication.


longquanzheng (approved):
Great! The title needs to be changed to be more informative before merge. It will become the commit message in main branch and release

PR #36, opened 2023-10-08
process id reuse policy

duoertai:


persistence/sql/sql_process_store.go
```diff
@@ -62,20 +63,92 @@ func (p sqlProcessStoreImpl) doStartProcessTx(
 	prcExeId := uuid.MustNewUUID()
 	hasNewWorkerTask := false
 
-	err := tx.InsertCurrentProcessExecution(ctx, extensions.CurrentProcessExecutionRow{
+	if req.ProcessStartConfig == nil {
```
longquanzheng, line 66: I don't feel like it's necessary to require this field. All the configs inside are "optional".

persistence/sql/sql_process_store.go
```diff
@@ -62,20 +63,92 @@ func (p sqlProcessStoreImpl) doStartProcessTx(
...
+		return nil, fmt.Errorf("processStartConfig is required")
+	}
+
+	prevProcessExecution, errGetPrevExecution := tx.SelectProcessExecutionForUpdate(ctx, prcExeId)
```
longquanzheng, line 70: This is a new executionId, there is no execution existing yet for this id.

longquanzheng: looks like you wanted to get the prevExecution, it should be after selectCurrentExecution

persistence/sql/sql_process_store.go
```diff
@@ -62,20 +63,92 @@ func (p sqlProcessStoreImpl) doStartProcessTx(
...
-	if err != nil {
-		if p.session.IsDupEntryError(err) {
-			// TODO support other ProcessIdReusePolicy on this error
-			return &persistence.StartProcessResponse{
-				AlreadyStarted: true,
-			}, nil
+	if errInsertCurrentProcessExecution != nil {
+		if p.session.IsDupEntryError(errInsertCurrentProcessExecution) {
+			// process with the same id is running
+			if req.ProcessStartConfig.GetIdReusePolicy() == xdbapi.TERMINATE_IF_RUNNING {
```
longquanzheng, line 81: it's better to put each policy handling in a smaller method

persistence/sql/sql_process_store.go
duoertai: default to reject_duplicate?

longquanzheng: In cadence the default is “failed only”(if previous exit abnormally). Then Temporal changed it to “allow if not running “. The reason behind that is that new users would just ask why their workflow cannot be started. It seems more friendly to new users. I would trust the decision from Temporal team for this default behavior

persistence/sql/sql_process_store.go
duoertai: if there is a running process with the same id, and id reuse policy is terminate_if_running, we still need the previous execution, right?

longquanzheng: Yes, to get the previous execution, we need to get the processExeucitonID from selectCurrentExecution. Doing it before selectCurrentExecution will have no value? 

Note that the processExeucitonID is uniquely created for every request.

extensions/tests/base_sql.go
```diff
@@ -0,0 +1,278 @@
+// Apache License 2.0
```
longquanzheng, line 1: I have refactored the test to https://github.com/xdblab/xdb/blob/main/persistence/sql/basetest to make it easier to maintain

extensions/postgres/transactional.go
```diff
@@ -5,27 +5,46 @@
...
 
-func (d dbTx) InsertCurrentProcessExecution(ctx context.Context, row extensions.CurrentProcessExecutionRow) error {
+func (d dbTx) InsertLatestProcessExecution(ctx context.Context, row extensions.LatestProcessExecutionRow) error {
 	row.ProcessExecutionIdString = row.ProcessExecutionId.String()
-	_, err := d.tx.ExecContext(ctx, insertCurrentProcessExecutionQuery, row.Namespace, row.ProcessId, row.ProcessExecutionIdString)
+	_, err := d.tx.ExecContext(ctx, insertLatestProcessExecutionQuery, row.Namespace, row.ProcessId, row.ProcessExecutionIdString)
+	return err
+}
+
+const selectLatestProcessExecutionForUpdateQuery = `SELECT namespace, process_id, process_execution_id FROM xdb_sys_latest_process_executions WHERE namespace=$1 AND process_id=$2 FOR UPDATE`
```
longquanzheng, line 34: For any new query/SQL interface, we should test in https://github.com/xdblab/xdb/blob/main/persistence/sql/basetest/base_basic.go 
(see my similar comment on https://github.com/xdblab/xdb/pull/44#discussion_r1359433331 )

engine/worker_task_conrrent_processor.go
```diff
@@ -186,7 +187,7 @@ func (w *workerTaskConcurrentProcessor) processWaitUntilTask(
 	}
 	if checkHttpError(err, httpResp) {
 		err := composeHttpError(err, httpResp)
-		w.logger.Info("worker API return error", tag.Error(err))
+		//w.logger.Info("worker API return error", tag.Error(err))
```
longquanzheng, line 190: curious why this is commented out?

duoertai: just for debug, sometimes there are too many logs so that I can't find the error log, will revert this change

extensions/postgres/transactional.go
```diff
@@ -15,17 +15,37 @@ package postgres
...
+
+const selectLatestProcessExecutionForUpdateQuery = `SELECT namespace, process_id, process_execution_id 
+FROM xdb_sys_latest_process_executions 
+WHERE namespace=$1 AND process_id=$2 FOR UPDATE`
+
+func (d dbTx) SelectLatestProcessExecutionForUpdate(
+	ctx context.Context, namespace string, processId string,
+) ([]extensions.LatestProcessExecutionRow, error) {
+	var rows []extensions.LatestProcessExecutionRow
+	err := d.tx.SelectContext(ctx, &rows, selectLatestProcessExecutionForUpdateQuery, namespace, processId)
```
longquanzheng, line 41: If it's expecting a single row, then use GetContext to ensure only single row is returned. https://jmoiron.github.io/sqlx/

persistence/sql/sql_process_store.go
```diff
@@ -91,28 +92,372 @@ func (p sqlProcessStoreImpl) StopProcess(ctx context.Context, request persistenc
 	return resp, err
 }
 
+func (p sqlProcessStoreImpl) markPendingStateAsAborted(
```
longquanzheng, line 95: After rebase, you will see that @zklgame has created a new way to mark the states aborted in a batch query, which is more efficient than the way that I did

persistence/sql/sql_process_store.go
```diff
@@ -91,28 +92,372 @@ func (p sqlProcessStoreImpl) StopProcess(ctx context.Context, request persistenc
...
 	req := request.Request
+
+	if req.ProcessStartConfig == nil {
+		req.ProcessStartConfig = &xdbapi.ProcessStartConfig{
+			IdReusePolicy: xdbapi.ALLOW_IF_NO_RUNNING.Ptr(),
+		}
+	}
+
+	if req.ProcessStartConfig.IdReusePolicy == nil {
+		req.ProcessStartConfig.IdReusePolicy = xdbapi.ALLOW_IF_NO_RUNNING.Ptr()
```
longquanzheng, line 155: this is dup code as three lines above.

persistence/sql/sql_process_store.go
```diff
@@ -91,28 +92,372 @@ func (p sqlProcessStoreImpl) StopProcess(ctx context.Context, request persistenc
...
+	}
+
+	switch *req.ProcessStartConfig.IdReusePolicy {
```
longquanzheng, line 158: nice! I like how it's organized here 👍

persistence/sql/sql_process_store.go
```diff
@@ -91,28 +92,372 @@ func (p sqlProcessStoreImpl) StopProcess(ctx context.Context, request persistenc
...
+		}
+
+		return &persistence.StartProcessResponse{
+			ProcessExecutionId: prcExeId,
+			AlreadyStarted:     false,
+			HasNewWorkerTask:   hasNewWorkerTask,
+		}, nil
+	}
+
+	prcExeId := uuid.MustNewUUID()
```
longquanzheng, line 259: nit: the below code can be abstracted in a separate method for reuse:
```
p.insertNewLatestProcessExecution(...)
```

persistence/sql/sql_process_store.go
```diff
@@ -91,28 +92,372 @@ func (p sqlProcessStoreImpl) StopProcess(ctx context.Context, request persistenc
...
 			}, nil
 		}
+
+		// if it is not running, but completed normally, return error
+		// otherwise, start a new process
+		if processExecutionRowForUpdate.Status == persistence.ProcessExecutionStatusCompleted {
+			return nil, fmt.Errorf("process %v has completed normally", request.Request.ProcessId)
+		}
+
+		prcExeId := uuid.MustNewUUID()
```
longquanzheng, line 313: similar here `p.UpdateLatestAndInsertNewProcessExecution`

persistence/sql/sqltest/base_basic.go
```diff
@@ -118,6 +118,245 @@ func SQLBasicTest(ass *assert.Assertions, store persistence.ProcessStore) {
...
+	// Start the process and verify it started correctly.
+	prcExeId := startProcess(ctx, ass, store, namespace, processId, input)
+
+	// Try to start the process again and verify the behavior.
+	retryStartProcessForFailure(ctx, ass, store, namespace, processId, input)
+
+	// Describe the process.
+	describeProcess(ctx, ass, store, namespace, processId, xdbapi.RUNNING)
+
+	// Test waitUntil API execution
```
longquanzheng, line 181: you probably don't have to test the worker task code path. I like the way the above test cases doing it to be more straightfoward

persistence/sql/sqltest/base_basic.go
```diff
@@ -118,6 +118,245 @@ func SQLBasicTest(ass *assert.Assertions, store persistence.ProcessStore) {
 	describeProcess(ctx, ass, store, namespace, processId, xdbapi.COMPLETED)
 }
 
+func SQLProcessIdReusePolicyAllowIfPreviousExitAbnormally(ass *assert.Assertions, store persistence.ProcessStore) {
```
longquanzheng, line 121: very clean and clear test case 👍

persistence/sql/sqltest/common.go
```diff
@@ -59,6 +59,150 @@ func startProcess(
...
+		Namespace:       namespace,
+		ProcessId:       processId,
+		ProcessStopType: xdbapi.TERMINATE,
+	})
+
+	ass.Nil(err)
+	ass.False(resp.NotExists)
+}
+
+func startProcessWithAllowIfPreviousExitAbnormally(ctx context.Context, ass *assert.Assertions,
```
longquanzheng, line 74: It seems like we are missing a test case for the default policy. Maybe add it into the test cases like `SQLProcessIdReusePolicyTerminateIfRunning` to see if it return error with the default policy

longquanzheng:
Very good job on the code quality/readability!

persistence/sql/sql_process_store.go
```diff
@@ -16,6 +16,8 @@ package sql
 import (
 	"context"
 	"fmt"
+	"time"
```
longquanzheng, line 19: this file is getting big. I think we can consider to split it based on API (you can do it in separate PR):
* sql_store_start_process.go
* sql_store_stop_process.go
* sql_store_common.go
cc @zklgame

extensions/postgres/transactional.go
duoertai: right, here I use selectContext is because there could be 0 or 1 row is returned, getContext will consider 0 row is an error, thus abort the transaction

persistence/sql/sqltest/base_basic.go
duoertai: this is to make the process complete and then start again

extensions/postgres/transactional.go
longquanzheng: I see. It’s probably better to return (Row, bool, error) and use the bool to indicate empty.

persistence/sql/sqltest/base_basic.go
longquanzheng: Ohh I see . Makes sense to me

persistence/sql/sql_process_store.go
duoertai: sure, I'll do it in the next PR

persistence/sql/sql_process_store.go
```diff
@@ -91,28 +92,372 @@ func (p sqlProcessStoreImpl) StopProcess(ctx context.Context, request persistenc
...
-		if p.session.IsDupEntryError(err) {
-			// TODO support other ProcessIdReusePolicy on this error
+		return nil, err
+	}
+	if len(latestProcessExecutions) > 1 {
+		return nil, fmt.Errorf("more than one running process execution for process id %v", request.Request.ProcessId)
+	}
+
+	if len(latestProcessExecutions) == 1 {
+		processExecutionRowForUpdate, err := tx.SelectProcessExecutionForUpdate(ctx, latestProcessExecutions[0].ProcessExecutionId)
```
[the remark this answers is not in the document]
longquanzheng: We need to add a other `SelectProcessExecution` (no ForUpdate) for ALLOW_IF_NO_RUNNING, ALLOW_IF_PREIVOUS_EXIT_ABNOARMALLY because they are not updating the current execution. This will save the locking contention.

longquanzheng (approved):
https://github.com/xdblab/xdb/pull/36#discussion_r1361431465
https://github.com/xdblab/xdb/pull/36#discussion_r1361419744 

lgtm, with two last comments ^

PR #39, opened 2023-10-10
Add timer queue and backoff retry for state API

longquanzheng:

Add backoff retry for state API

## What has changed

- API changes: https://github.com/xdblab/xdb-apis/pull/6
- AsyncService: Add timer queue, timer queue concurrent processor
- Config: Add timer queue config
- ProcessStore: Add timer queue APIs
- postgres: add timer queue implementation

* Integration test https://github.com/xdblab/xdb-golang-sdk/pull/13
* Persistence test : TODO


longquanzheng:
I am still working on adding more tests. Will appreciate any early review feedback!

extensions/postgres/non_transactional.go
```diff
@@ -72,3 +72,24 @@ func (d dbSession) BatchDeleteWorkerTask(
 	_, err := d.db.ExecContext(ctx, batchDeleteWorkerTaskQuery, filter.ShardId, filter.MinTaskSequenceInclusive, filter.MaxTaskSequenceInclusive)
 	return err
 }
+
+const batchSelectTimerTasksOfFirstPageQuery = `SELECT 
+    shard_id, fire_timestamp, task_sequence, process_execution_id, state_id, state_id_sequence, task_type, info
+	FROM xdb_sys_timer_tasks WHERE shard_id = $1 AND fire_timestamp>=$2 AND task_sequence>= $3 ORDER BY task_sequence ASC LIMIT $4`
+
+func (d dbSession) BatchSelectTimerTasks(ctx context.Context, filter extensions.TimerTaskRangeSelectFilter) ([]extensions.TimerTaskRow, error) {
```
zklgame, line 80: I am not sure whether this query can return the desired data. E.g., if the table contains:
```
task_sequence, fire_timestamp
1, 3
2, 2
3, 1
```
And limit is 2. Then we should get the following rows ordering by the fire_timestamp?
```
3, 1
2, 2
```

zklgame, line 80: I am not sure whether this query can return the desired data. E.g., if the table contains:
```
task_sequence, fire_timestamp
1, 3
2, 2
3, 1
```
And limit is 2. Then we should get the following rows ordering by the fire_timestamp?
```
3, 1
2, 2
```

zklgame, line 80: I am not sure whether this query can return the desired data. E.g., if the table contains:
```
task_sequence, fire_timestamp
1, 3
2, 2
3, 1
```
And limit is 2. Then we should get the following rows ordering by the fire_timestamp?
```
3, 1
2, 2
```

engine/timer_task_conrrent_processor.go
```diff
@@ -0,0 +1,149 @@
...
+				case task, ok := <-w.taskToProcessChan:
+					if !ok {
+						return
+					}
+					if !w.currentShards[task.ShardId] {
+						w.logger.Info("skip the stale task that is due to shard movement", tag.Shard(task.ShardId), tag.ID(task.GetId()))
+						continue
+					}
+
+					err := w.processTimerTask(task)
```
zklgame, line 90: How could we make sure the task.fire_timestamp <= now()?

engine/timer_task_queue.go
```diff
@@ -0,0 +1,194 @@
...
+			page := &timerTaskPage{
+				minTaskIndex: minIndex,
+				maxTaskIndex: maxIndex,
+				pendingCount: len(resp.Tasks),
+			}
+
+			for _, task := range resp.Tasks {
+				w.processor.GetTasksToProcessChan() <- task
+				index := newTimerIndex(task.FireTimestampSeconds, *task.TaskSequence)
+				w.pendingTaskIndexToPage[index] = page
```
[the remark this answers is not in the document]
longquanzheng: Yeah that's true. I will update it to be simpler

extensions/postgres/non_transactional.go
longquanzheng: oops. I copied paste the query and forgot to fix the order. It should be `ORDER BY fire_timestamp , task_sequence ASC`

engine/timer_task_conrrent_processor.go
longquanzheng: 🤦 I will fix it

extensions/postgres/tests/postgres_test.go
```diff
@@ -22,45 +22,19 @@
 package tests
 
 import (
-	"fmt"
```
longquanzheng, line 25: @zklgame @duoertai This should have been done this earlier so that it's easier to add more persistence test. I am going to make another PR to do this separately, so that you can add test based on this pattern

persistence/sql/cmmon.go
```diff
@@ -0,0 +1,57 @@
...
+		})
+	}
+	resp := &persistence.GetTimerTasksResponse{
+		Tasks: tasks,
+	}
+	if len(dbTimerTasks) > 0 {
+		firstTask := dbTimerTasks[0]
+		lastTask := dbTimerTasks[len(dbTimerTasks)-1]
+		resp.MinFireTimestampSecondsInclusive = firstTask.FireTimeUnixSeconds
+		resp.MinSequenceInclusive = firstTask.TaskSequence
```
zklgame, line 52: This logic seems not correct to me since a later task can have a smaller fire time.

engine/timer_task_queue.go
```diff
@@ -0,0 +1,295 @@
...
+		// The next preload will poll them anyway
+	} else {
+		if len(resp.Tasks) > 0 {
+			// update the max loaded sequence so that next time it won't load the same tasks
+			// currMaxLoadedTaskTimestamp is not updated because the new tasks won't have a bigger timestamp
+			w.currMaxLoadedTaskSequence = resp.MaxSequenceInclusive
+
+			// add the new tasks into the heap
+			for _, task := range resp.Tasks {
+				heap.Push(&w.remainingToFireTimersHeap, &task)
```
zklgame, line 268: Will it push duplicated tasks into the heap?

persistence/sql/cmmon.go
longquanzheng: In the select statement, we always have `ORDER BY fire_time_unix_seconds, task_sequence ASC` in the PR (see non_transactional.go`. 

Maybe I can change the name to `SelectTimerTasksForTimestampsOrdered` and `BatchSelectTimerTasksOrdered` to be more clear

extensions/postgres/non_transactional.go
```diff
@@ -72,3 +73,42 @@ func (d dbSession) BatchDeleteWorkerTask(
...
+}
+
+// const selectTimerTasksForTimestampsQuery = `SELECT
+//
+//	   shard_id, fire_time_unix_seconds, task_sequence, process_execution_id, state_id, state_id_sequence, task_type, info
+//		FROM xdb_sys_timer_tasks WHERE shard_id = %v AND fire_time_unix_seconds IN (%v) AND task_sequence <= %v
+//		ORDER BY fire_time_unix_seconds, task_sequence ASC LIMIT %v`
+const selectTimerTasksForTimestampsQuery = `SELECT 
+    shard_id, fire_time_unix_seconds, task_sequence, process_execution_id, state_id, state_id_sequence, task_type, info
+	FROM xdb_sys_timer_tasks WHERE shard_id = :shard_id AND fire_time_unix_seconds IN (:fire_time_unix_seconds) AND task_sequence <= :min_task_sequence_inclusive
```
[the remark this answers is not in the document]
longquanzheng: ahh right. good catch

engine/timer_task_queue.go
longquanzheng: that's the idea of tracking `currMaxLoadedTaskSequence`, and use it for every `GetTimerTasksForTimestamps`. Because the taskSequence is always increasing meaning the new tasks will also have a higher sequence number. The dup tasks will be filtered out when selecting it.

extensions/postgres/non_transactional.go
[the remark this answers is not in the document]
longquanzheng: actually, I should changed "<=" to ">=" 🤦 It's meant to load the tasks with at least this task sequence

engine/timer_task_queue.go
```diff
@@ -0,0 +1,295 @@
...
+			}
+
+			minTimestamp := time.Unix(resp.MinFireTimestampSecondsInclusive, 0)
+			if !w.nextFiringTimer.IsActive() || w.nextFiringTimer.FireAfter(minTimestamp) {
```
zklgame, line 272: What if the nextFiringTimer is active but the nextFiringTimer > minTimestamp?

engine/timer_task_queue.go
```diff
@@ -0,0 +1,295 @@
...
+			select {
+			case <-w.nextPreloadTimer.FireChan():
+				if w.shouldLoadNextBatch() {
+					w.loadAndDispatchAndPrepareNext()
+				}
+			case <-w.nextFiringTimer.FireChan():
+				w.sendFiredTimerToProcessor()
+			case task, ok := <-w.tasksCompletionChan:
+				if ok {
+					delete(w.firedToCompleteTimerSequenceMap, *task.TaskSequence)
```
zklgame, line 133: I didn't find anywhere this `firedToCompleteTimerSequenceMap` is used.

engine/timer_task_queue.go
```diff
@@ -0,0 +1,295 @@
...
+	w.nextPreloadTimer.Update(time.Now()) // fire immediately to make the first poll
+
+	go func() {
+		for {
+			select {
+			case <-w.nextPreloadTimer.FireChan():
+				if w.shouldLoadNextBatch() {
+					w.loadAndDispatchAndPrepareNext()
+				}
+			case <-w.nextFiringTimer.FireChan():
```
zklgame, line 129: Besides firing the heap[0] after loadAndDispatch,  where is the logic to contiously fire other tasks in the heap?

engine/timer_task_queue.go
longquanzheng: `nextFiringTimer > minTimestamp` means `w.nextFiringTimer.FireAfter(minTimestamp)`,  then the `nextFiringTimer` will be updated and fire with the `minTimestamp` instead. 

If `nextFiringTimer <= minTimestamp`, then the `nextFiringTimer` doesn't need to be updated. because it's already earlier than minTimestamp.

engine/timer_task_queue.go
longquanzheng: In `shouldLoadNextBatch()`, it's used to check if all the fired timers have been completed.

engine/timer_task_queue.go
longquanzheng: in `sendFiredTimerToProcessor()`, there is a loop which continuously pop the item if it's less than Now().

engine/timer_task_queue.go
zklgame: I couldn't locate where any element was added to the firedToCompleteTimerSequenceMap.




 was put into firedToCompleteTimerSequenceMap.

zklgame: I think we actually don't need `firedToCompleteTimerSequenceMap` because the fired timer_task just becomes worker_task.

zklgame (approved):
Great implementation!

engine/timer_task_queue.go
longquanzheng: ohh yeah you are right. Originally I wanted to use that to make sure it's safe for timerqueue to delete the fired timers.
But l found that it's better to let the processor delete when processing (otherwise, timer will be double firing and it's hard to handle). 
So now we can get rid of `firedToCompleteTimerSequenceMap`. In the future, for other timer tasks, we can just follow the same way to process+delete in an transaction.

PR #42, opened 2023-10-11
Add StopProcess

zklgame:

Add `StopProcess`.

## What has changed

- update tables (see `doStopProcessTx`).

https://github.com/xdblab/xdb-golang-sdk/pull/14


extensions/postgres/non_transactional.go
```diff
@@ -35,6 +36,14 @@ func (d dbSession) SelectCurrentProcessExecution(
 	return &row, err
 }
 
+const deleteCurrentProcessExecutionQuery = `DELETE
```
longquanzheng, line 39: For stopProcess API, we actually don't want to delete anything. The deletion will be done by the retention policy feature. Similar to Cadence/Temporal, all the history/execution records will be deleted after rention time beyond the execution is closed. 

However, this query is useful, you can keep it if you want

zklgame: True.

extensions/postgres/non_transactional.go
```diff
@@ -72,3 +73,21 @@ func (d dbSession) BatchDeleteWorkerTask(
 	_, err := d.db.ExecContext(ctx, batchDeleteWorkerTaskQuery, filter.ShardId, filter.MinTaskSequenceInclusive, filter.MaxTaskSequenceInclusive)
 	return err
 }
+
+const deleteWorkerTasksQuery = `DELETE
+	FROM xdb_sys_worker_tasks where process_execution_id = $1
+`
+
+func (d dbSession) DeleteWorkerTasks(ctx context.Context, processExecutionId uuid.UUID) error {
```
zklgame, line 81: @longquanzheng Which case will the system have a better performance when stopping process execution?

- detele the related worker/timer rows in db so that they will not put into process queue, thus we can save queue storage.
- do nothing so the tasks/timers will go to the process queue but do nothing during execution.

extensions/postgres/transactional.go
```diff
@@ -106,6 +106,34 @@ func (d dbTx) UpdateAsyncStateExecution(
...
+WHERE process_execution_id=$1 AND wait_until_status=1
+`
+
+const updateAsyncStateExecutionToAbortRunningQuery = `UPDATE xdb_sys_async_state_executions set
+version = version + 1,
+execute_status = 5
+WHERE process_execution_id=$1 AND execute_status=1
+`
+
+func (d dbTx) UpdateAsyncStateExecutionToAbortRunning(
```
zklgame, line 121: I see another implementation is [here](https://github.com/xdblab/xdb/blob/e8b9ad0c202d80367a380c0ad218a5ea64defa64/persistence/sql/sql_process_store.go#L451).

extensions/postgres/transactional.go
```diff
@@ -106,6 +106,34 @@ func (d dbTx) UpdateAsyncStateExecution(
 	return nil
 }
 
+const updateAsyncStateWaitUntilToAbortRunningQuery = `UPDATE xdb_sys_async_state_executions set
```
longquanzheng, line 109: Nit: maybe call it “batchUpdate…” to be more clear

zklgame:
@longquanzheng this MR is ready now.

persistence/sql/sql_process_store.go
```diff
@@ -159,6 +182,61 @@ func (p sqlProcessStoreImpl) doStartProcessTx(
...
+
+	err = tx.UpdateProcessExecution(ctx, *procExecRow)
+	if err != nil {
+		return nil, err
+	}
+
+	// handle xdb_sys_async_state_executions
+	// find all related rows with the processExecutionId, and
+	// modify the wait_until/execute status from running to aborted
+	err = tx.UpdateAsyncStateExecutionToAbortRunning(ctx, curProcExecRow.ProcessExecutionId)
```
longquanzheng, line 230: Minor optimization idea: since we have loaded `PendingExecutionMap`, it would be better to check if there is any pending executions -- skip this update if there is no. 

It is possible that the process execution is still running when there is no state execution is running(because RPC can still trigger new state execution later)

longquanzheng (approved):
very good job!!!

extensions/postgres/transactional.go
```diff
@@ -106,6 +106,34 @@ func (d dbTx) UpdateAsyncStateExecution(
...
+WHERE process_execution_id=$1 AND wait_until_status=1
+`
+
+const updateAsyncStateExecutionToAbortRunningQuery = `UPDATE xdb_sys_async_state_executions set
+version = version + 1
+execute_status = 5
+WHERE process_execution_id=$1 AND execute_status=1
+`
+
+func (d dbTx) UpdateAsyncStateExecutionToAbortRunning(
```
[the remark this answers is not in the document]
zklgame: This one looks good to me:
```
UPDATE xdb_sys_async_state_executions SET
version = CASE WHEN wait_until_status=1 OR execute_status=1 THEN version + 1 ELSE version END,
wait_until_status = CASE WHEN wait_until_status=1 THEN 5 ELSE wait_until_status END,
execute_status = CASE WHEN execute_status=1 THEN 5 ELSE execute_status END
WHERE process_execution_id=$1
```

persistence/sql/sql_process_store.go
zklgame: Good idea!

PR #44, opened 2023-10-14
Add graceful complete and force fail

zklgame:

Support GracefulComplete and ForceFail StateDecision.

## What has changed

- add `wait_to_complete` field to table `xdb_sys_process_executions`
- add `SELECT FOR UPDATE` lock to `selectProcessExecutionForUpdateQuery`
- add support for GRACEFUL_COMPLETE_PROCESS and FORCE_FAIL_PROCESS

https://github.com/xdblab/xdb-golang-sdk/pull/15


extensions/postgres/schema/xdb_sys_schema.sql
```diff
@@ -16,7 +16,7 @@ CREATE TABLE xdb_sys_process_executions(
     timeout_seconds INTEGER,
     history_event_id_sequence INTEGER,
     state_execution_sequence_maps jsonb NOT NULL , -- some maps from stateId and sequence number
-    info jsonb , -- workerURL, processType, etc
+    info jsonb , -- workerURL, processType, waitToComplete, etc
```
longquanzheng, line 19: It will be slightly better perf to use a separate bool field, as updating a blob will be more cost. 
so usually, I would prefer using this info field to store static info only. (And that’s why other info like sequenceMap is in separate field) 

However , not that this also because of this stage of the project. This blob field can be used to add new fields without schema change. After we make a production release, we may choose to use it for adding even a bool field so that users don’t have to change schema in order to upgrade

extensions/postgres/transactional.go
```diff
@@ -69,6 +70,17 @@ func (d dbTx) SelectAsyncStateExecutionForUpdate(
 	return &row, err
 }
 
+const selectRunningAsyncStateExecutionCountQuery = `SELECT
```
longquanzheng, line 73: I think the pending sequence map in the process execution can provide the information

extensions/sql_db_interfaces.go
```diff
@@ -55,9 +55,11 @@ type transactionalCRUD interface {
 	UpdateProcessExecution(ctx context.Context, row ProcessExecutionRowForUpdate) error
 
 	SelectAsyncStateExecutionForUpdate(ctx context.Context, filter AsyncStateExecutionSelectFilter) (*AsyncStateExecutionRow, error)
+	SelectRunningAsyncStateExecutionCount(ctx context.Context, processExecutionId uuid.UUID) (int, error)
```
longquanzheng, line 58: I forgot to mention it last PR, it’s usually easier to test the code, if you adding/change a test case in postgres_test.go to just verify the database code path is working as expected. Based on that, testing other engine and service layer would be much easier

extensions/postgres/schema/xdb_sys_schema.sql
zklgame: Learnt!

persistence/sql/basetest/common.go
```diff
@@ -35,6 +35,13 @@ func createTestInput() xdbapi.EncodedObject {
 	}
 }
 
+func createNilInput() xdbapi.EncodedObject {
```
longquanzheng, line 38: Nit: createEmotyEncodedObject

longquanzheng (approved):
Lgtm. It would be better to rebase to main branch and add some new test case for the completeExecute persistence api. The latest commit in main branch has made it simpler to add new persistence test case (see my msg in WeChat )

zklgame:
> Lgtm. It would be better to rebase to main branch and add some new test case for the completeExecute persistence api. The latest commit in main branch has made it simpler to add new persistence test case (see my msg in WeChat )

Added two more tests based on my best understanding. Though I think each test needs a long setup. Maybe we need to further refactor the testing codes to make a common setup.

longquanzheng:
> > Lgtm. It would be better to rebase to main branch and add some new test case for the completeExecute persistence api. The latest commit in main branch has made it simpler to add new persistence test case (see my msg in WeChat )
> 
> Added two more tests based on my best understanding. Though I think each test needs a long setup. Maybe we need to further refactor the testing codes to make a common setup.

I think the way “testMain” works is that it will reuse the same setup. You can see that it will only create test database once

zklgame:
> > > Lgtm. It would be better to rebase to main branch and add some new test case for the completeExecute persistence api. The latest commit in main branch has made it simpler to add new persistence test case (see my msg in WeChat )
> > 
> > 
> > Added two more tests based on my best understanding. Though I think each test needs a long setup. Maybe we need to further refactor the testing codes to make a common setup.
> 
> I think the way “testMain” works is that it will reuse the same setup. You can see that it will only create test database once

The setup I am talking about is start process and completeExecute to run other states. Now for each test I need to do the same operations. So I think we can put the start process and start other states operations in a common shared logic?

zklgame:
The integ test failed after merging due to https://github.com/xdblab/xdb-golang-sdk/pull/17.

PR #54, opened 2023-10-19
Merge intervals when committing completed pages

zklgame:

Merge intervals when committing completed pages

## What has changed

- API change: ...
- Golang SDK change: ...
- Server Component 1: ...
- Server Component 2: ...

## Checklist before merge

### DIFF
--- a/engine/worker_task_queue.go
+++ b/engine/worker_task_queue.go
@@ -165,23 +165,7 @@ func (w *workerTaskQueueImpl) pollAndDispatchAndPrepareNext() {
 
 func (w *workerTaskQueueImpl) commitCompletedPages(ctx context.Context) error {
 	if len(w.completedPages) > 0 {
-		// merge pages, e.g.,
-		// [1, 2], [3, 4], [7, 8] -> [1, 4], [7, 8]
-		sort.Slice(w.completedPages, func(i, j int) bool {
-			return w.completedPages[i].minTaskSequence < w.completedPages[j].minTaskSequence
-		})
-		var pages []*workerTaskPage
-		for _, page := range w.completedPages {
-			if len(pages) == 0 || pages[len(pages)-1].maxTaskSequence+1 < page.minTaskSequence {
-				pages = append(pages, page)
-			} else {
-				if pages[len(pages)-1].maxTaskSequence < page.maxTaskSequence {
-					pages[len(pages)-1].maxTaskSequence = page.maxTaskSequence
-				}
-			}
-		}
-
-		w.completedPages = pages
+		w.completedPages = mergeWorkerTaskPages(w.completedPages)
 
 		for idx, page := range w.completedPages {
 			req := persistence.DeleteWorkerTasksRequest{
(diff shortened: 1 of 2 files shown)


engine/worker_task_queue.go
```diff
@@ -164,7 +165,24 @@ func (w *workerTaskQueueImpl) pollAndDispatchAndPrepareNext() {
...
+			if len(pages) == 0 || pages[len(pages)-1].maxTaskSequence+1 < page.minTaskSequence {
+				pages = append(pages, page)
+			} else {
+				if pages[len(pages)-1].maxTaskSequence < page.maxTaskSequence {
+					pages[len(pages)-1].maxTaskSequence = page.maxTaskSequence
+				}
+			}
+		}
+
+		w.completedPages = pages
```
zklgame, line 184: @longquanzheng Do we need to add lock on `w.completedPages` when commiting the completed pages? I feel `receiveCompletedTask` will modify the `w.completedPages` at the same time?

longquanzheng: oh we don't need to. It's a single thread component. WorkerTaskProcessor is the concurrent component that will have multiple threads. 
When this is deleting, the msg sent to the channel won't be consumed/processed

engine/worker_task_queue.go
```diff
@@ -164,7 +165,24 @@ func (w *workerTaskQueueImpl) pollAndDispatchAndPrepareNext() {
 
 func (w *workerTaskQueueImpl) commitCompletedPages(ctx context.Context) error {
 	if len(w.completedPages) > 0 {
-		// TODO optimize by combining all pages into single query (as long as it won't exceed certain limit)
+		// merge pages, e.g.,
+		// [1, 2], [3, 4], [7, 8] -> [1, 4], [7, 8]
```
longquanzheng, line 169: Nit: maybe put this code in a separate method

engine/worker_task_queue.go
```diff
@@ -164,7 +165,24 @@ func (w *workerTaskQueueImpl) pollAndDispatchAndPrepareNext() {
...
+				pages = append(pages, page)
+			} else {
+				if pages[len(pages)-1].maxTaskSequence < page.maxTaskSequence {
+					pages[len(pages)-1].maxTaskSequence = page.maxTaskSequence
+				}
+			}
+		}
+
+		w.completedPages = pages
+
```
longquanzheng, line 185: In addition to merge interval, I was thinking that the deleting persistnce API can take a list of intervals to delete them in a single query. But you can also do it separately.

longquanzheng (approved):

engine/worker_task_queue.go
longquanzheng: also it's better to have a "unit test" for this

engine/worker_task_queue.go
```diff
@@ -164,7 +165,24 @@ func (w *workerTaskQueueImpl) pollAndDispatchAndPrepareNext() {
...
+		sort.Slice(w.completedPages, func(i, j int) bool {
+			return w.completedPages[i].minTaskSequence < w.completedPages[j].minTaskSequence
+		})
+		var pages []*workerTaskPage
+		for _, page := range w.completedPages {
+			if len(pages) == 0 || pages[len(pages)-1].maxTaskSequence+1 < page.minTaskSequence {
+				pages = append(pages, page)
+			} else {
+				if pages[len(pages)-1].maxTaskSequence < page.maxTaskSequence {
+					pages[len(pages)-1].maxTaskSequence = page.maxTaskSequence
```
longquanzheng, line 179: It may be better to create new pages rather than modify the existing

engine/worker_task_queue.go
zklgame: Oh I forgot to do the sql-side. Later I did some investigations but found it cumbersome to do so. One possible solutions is to generate all task_sequence from the intervals and let the sql `delete from xxx in (..)`. But I think this is not ideal way.

engine/worker_task_queue.go
zklgame: Yeah this is better.

engine/worker_task_queue.go
zklgame: Learnt!

PR #55, opened 2023-10-19
fix deadlock by locking proc exec row first

duoertai:

## What has changed

- API change: ...
- Golang SDK change: ...
- Server Component 1: ...
- Server Component 2: ...

## Checklist before merge

### DIFF
--- a/persistence/sql/sql_process_store.go
+++ b/persistence/sql/sql_process_store.go
@@ -269,6 +269,12 @@ func (p sqlProcessStoreImpl) applyTerminateIfRunningPolicy(
 		return nil, err
 	}
 
+	// lock process execution row first
+	_, err = tx.SelectProcessExecutionForUpdate(ctx, latestProcessExecution.ProcessExecutionId)
+	if err != nil {
+		return nil, err
+	}
+
 	// if it is still running, terminate it and start a new process
 	// otherwise, start a new process
 	if found {
@@ -459,7 +465,6 @@ func (p sqlProcessStoreImpl) doStopProcessTx(
 	// handle xdb_sys_process_executions
 	procExecRow, err := tx.SelectProcessExecutionForUpdate(ctx, curProcExecRow.ProcessExecutionId)
 	if err != nil {
-		p.logger.Error(err.Error())
 		return nil, err
 	}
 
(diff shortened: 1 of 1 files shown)


persistence/sql/sql_process_store.go
```diff
@@ -269,6 +269,12 @@ func (p sqlProcessStoreImpl) applyTerminateIfRunningPolicy(
 		return nil, err
 	}
 
+	// lock process execution row first
+	_, err = tx.SelectProcessExecutionForUpdate(ctx, latestProcessExecution.ProcessExecutionId)
```
longquanzheng, line 273: How about just replace line 281 with this? 
Because the else(not found) branch won't update anything for this execution so it's still a bit wasteful to lock it.

persistence/sql/sql_process_store.go
```diff
@@ -658,7 +662,13 @@ func (p sqlProcessStoreImpl) doCompleteWaitUntilExecutionTx(
 		LastFailure:        nil,
 	}
 
-	err := tx.UpdateAsyncStateExecution(ctx, stateRow)
+	// lock process execution row first
```
longquanzheng, line 665: Actually, maybe just not lock it for now. Because this method is not updating the process execution at all. 
Can you just leave a comment here for the future: if this method is going to update process execution, it should lock it before updating or locking the async state executions. 

We will probably need to do that very soon, when @zklgame  is adding the localQueue command. But we may not need to update processExe for timerCommand, or no command.

longquanzheng (changes requested):
can you also enable the integ test that I commented out because of the deadlock?

persistence/sql/sql_process_store.go
duoertai: right, not found case is just creating states for a new process

persistence/sql/sql_process_store.go
duoertai: yep, I noticed this too, but waitUntil will need to do it

duoertai: should we just leave it there?

longquanzheng (approved):

PR #61, opened 2023-10-20
Rename workerXXX to immediateXXX

zklgame:

Rename most workerXXX to immediateXXX, except the `WorkerTaskBackoffRetry` definitions.


extensions/postgres/schema/xdb_sys_schema.sql
```diff
@@ -37,15 +37,23 @@ CREATE TABLE xdb_sys_async_state_executions(
    PRIMARY KEY (process_execution_id, state_id, state_id_sequence)
 );
 
-CREATE TABLE xdb_sys_worker_tasks(
+CREATE TABLE xdb_sys_immediate_tasks(
     shard_id INTEGER NOT NULL, -- for sharding xdb-local-mq
     task_sequence bigserial,   
     --
-    task_type SMALLINT NOT NULL, -- 1: waitUntil 2: execute
-    process_execution_id uuid, -- for looking up xdb_sys_async_state_executions
```
zklgame, line 45: Here is the key schema modification to support local_queue_message by reusing the woker_tasks table.

longquanzheng: I feel like it's better to just rename the table as well. xdb_sys_worker_tasks ==> xdb_sys_immediate_tasks.

persistence/data_models.go
```diff
@@ -51,51 +51,51 @@ type (
...
 		ShardId int32
 		// TaskSequence represents the increasing order in the queue of the shard
 		// It should be empty when inserting, because the persistence/database will
 		// generate the value automatically
 		TaskSequence *int64
 
-		TaskType WorkerTaskType
+		TaskType ImmediateTaskType
 
-		ProcessExecutionId uuid.UUID
```
longquanzheng, line 96: I don't think this needs to be renamed. 
It's to tell which processExecutionId the task is for. E.g. for LocalQueueMessageReceived task, we also need a process exe Id.

persistence/data_models_enums.go
```diff
@@ -78,19 +78,22 @@ func (e StateExecutionStatus) String() string {
...
 
-type WorkerTaskType int32
+type ImmediateTaskType int32
 
 const (
-	WorkerTaskTypeWaitUntil WorkerTaskType = 1
-	WorkerTaskTypeExecute   WorkerTaskType = 2
+	ImmediateTaskTypeWaitUntil         ImmediateTaskType = 1
+	ImmediateTaskTypeExecute           ImmediateTaskType = 2
+	ImmediateTaskTypeLocalQueueMessage ImmediateTaskType = 3
```
longquanzheng, line 86: Maybe `ImmediateTaskTypeLocalQueueMessageProcessing` to be more clear.

extensions/postgres/schema/xdb_sys_schema.sql
zklgame: Already renamed.

persistence/data_models.go
zklgame: For LocalQueueMessageReceived, I think we just need to store the dedupId, then we can get the sent message from the local_queue table by the dedupId.

extensions/postgres/schema/xdb_sys_schema.sql
longquanzheng: ahh right. I misread it

extensions/postgres/schema/xdb_sys_schema.sql
```diff
@@ -37,15 +37,23 @@ CREATE TABLE xdb_sys_async_state_executions(
...
+    task_type SMALLINT NOT NULL, -- 1: waitUntil 2: execute 3: localQueueMessage
+    -- the uuid represents various UUIDs depending on the `task_type`:
+    -- if the `task_type` is waitUntil or execute, the value corresponds to the `process_execution_id`.
+    -- if the `task_type` is localQueueMessage, the value corresponds to the message `dedup_id`.
+    uuid uuid,
```
longquanzheng, line 48: localQueueMessage actually also need a process exection id as well. 

Maybe put the dedup_id into info field, also add a new field

persistence/data_models.go
longquanzheng: ah right, we will need both the PK in local_queue table is <proc_exe_id, queueName, dedup_id> based on our doc :P

extensions/postgres/schema/xdb_sys_schema.sql
longquanzheng: Related to the comment xdb-apis PR, a single task could be for multiple messages -- so it's more reasonable to me that the dedupId should live in the info field. 

I think keep prc_exe_id is quite important -- for a very long time, I can for see that a task is always associated with a process execution. Some tasks may not have state_id/state_id_seq, like the localMsgReceived that you are adding. but process_execution_id will be a critical part for any tasks. 

In the very long future, when we started to design "global queue", we may need to think about tasks that associated with more than one process executions(as global queue is shared acorss proc exes). At that time, I guess this proc exe could be empty.

persistence/data_models.go
zklgame: keep process_execution_id in the woker table acutally can save one query to get the process_execution_id from local_queue table. Yeah it's worth adding it back.

extensions/postgres/schema/xdb_sys_schema.sql
zklgame: Good idea. I think we can put the `queue_name`, `dedup_id` and `payload` of a message into info, so that we don't need to add another field in the table.

persistence/data_models_enums.go
zklgame: Chosed `ImmediateTaskTypeNewLocalQueueMessage`.

persistence/sql/sql_process_store.go
```diff
@@ -50,7 +50,7 @@ func (p sqlProcessStoreImpl) Close() error {
 	return p.session.Close()
 }
 
-func (p sqlProcessStoreImpl) CleanUpTasksForTest(ctx context.Context, shardId int32) error {
+func (p sqlProcessStoreImpl) CleanUpImmediateTasksForTest(ctx context.Context, shardId int32) error {
```
longquanzheng, line 53: Looks like a typo?

longquanzheng (approved):
excellent!

persistence/data_models_enums.go
```diff
@@ -78,19 +78,22 @@ func (e StateExecutionStatus) String() string {
...
 
-type WorkerTaskType int32
+type ImmediateTaskType int32
 
 const (
-	WorkerTaskTypeWaitUntil WorkerTaskType = 1
-	WorkerTaskTypeExecute   WorkerTaskType = 2
+	ImmediateTaskTypeWaitUntil            ImmediateTaskType = 1
+	ImmediateTaskTypeExecute              ImmediateTaskType = 2
+	ImmediateTaskTypeNewLocalQueueMessage ImmediateTaskType = 3
```
longquanzheng, line 86: Probably plural to be more accurate : `ImmediateTaskTypeNewLocalQueueMessages`

persistence/data_models_json_fields.go
```diff
@@ -137,23 +137,26 @@ type WorkerTaskBackoffInfoJson struct {
 	FirstAttemptTimestampSeconds int64 `json:"firstAttemptTimestampSeconds"`
 }
 
-type WorkerTaskInfoJson struct {
+type ImmediateTaskInfoJson struct {
+	// used when the `task_type` is waitUntil or execute
 	WorkerTaskBackoffInfo *WorkerTaskBackoffInfoJson `json:"workerTaskBackoffInfo"`
+	// used when the `task_type` is localQueueMessage
+	LocalQueueMessagePayload *xdbapi.EncodedObject `json:"localQueueMessagePayload"`
```
longquanzheng, line 144: I think this refactor is big enough for a PR. It's not too late to do it later when you are implementing the feature -- I feel like we may not need the payload. Probably dedupIds/queueNames are enough if necessary.

persistence/sql/sql_process_store.go
zklgame: Oh yeah I thought this method only cleans up the immediate_task table. After checking I found this method cleans up both task tables. Renamed it back to `CleanUpTasksForTest `.

persistence/data_models_enums.go
zklgame: True.

persistence/data_models_json_fields.go
zklgame: True. I modifed this field to be :
```
type LocalQueueMessageInfoJson struct {
	QueueName string    `json:"queueName"`
	DedupId   uuid.UUID `json:"dedupId"`
}
```

persistence/data_models_enums.go
zklgame: Wait, I think one row in local_queue table corresponds to one row in task table. So when client send 2 messages to local queue, we will insert two rows into local_queue table and insert two rows into task table.

If this is correct understanding, then the task type should be in a singular form?

zklgame: I will leave it in a singular form later. If we are going to store multiple messages in a row, we can rename it to the plural form.

PR #63, opened 2023-10-21
Add local queue message + command request support

zklgame:

- Implement local queue message, support ANY_OF_COMPLETION and ALL_OF_COMPLETION of LocalQueueCommand.
- Merge wait_until_status and execute_status into one status

## What has changed

- implement `publish to local queue` and `process local queue messages task` (batch process).
- completeWaitUntil -> processWaitUntil. processWaitUntil is later divided into complete or update methods.
- implement ANY_OF_COMPLETION and ALL_OF_COMPLETION of LocalQueueCommand.
- Merge wait_until_status and execute_status into one status

https://github.com/xdblab/xdb-java-sdk/pull/9

## Checklist before merge


extensions/postgres/schema/xdb_sys_schema.sql
```diff
@@ -15,6 +15,7 @@ CREATE TABLE xdb_sys_process_executions(
     timeout_seconds INTEGER,
     history_event_id_sequence INTEGER,
     state_execution_sequence_maps jsonb NOT NULL , -- some maps from stateId and sequence number
+    state_execution_waiting_queues jsonb, -- { queue_1: [(state_id, state_id_sequence, needed_count), ...], }
```
zklgame, line 18: This field will be used to expedite the process of locating the target state for message consumption.

persistence/data_models_enums.go
```diff
@@ -48,13 +48,14 @@ func (e ProcessExecutionStatus) String() string {
...
-	StateExecutionStatusTimeout   StateExecutionStatus = 4
-	StateExecutionStatusAborted   StateExecutionStatus = 5
+	StateExecutionStatusSkipped            StateExecutionStatus = -1
+	StateExecutionStatusUndefined          StateExecutionStatus = 0
+	StateExecutionStatusRunning            StateExecutionStatus = 1
+	StateExecutionStatusCompleted          StateExecutionStatus = 2
+	StateExecutionStatusFailed             StateExecutionStatus = 3
+	StateExecutionStatusTimeout            StateExecutionStatus = 4
+	StateExecutionStatusAborted            StateExecutionStatus = 5
+	StateExecutionStatusWaitingForCommands StateExecutionStatus = 6
```
zklgame, line 58: Need a new status to prevent repeatedly running wait_until tasks that haven't been deleted.

extensions/postgres/schema/xdb_sys_schema.sql
```diff
@@ -64,4 +65,12 @@ CREATE TABLE xdb_sys_timer_tasks(
     state_id_sequence INTEGER, -- for looking up xdb_sys_async_state_executions
     info jsonb ,
     PRIMARY KEY (shard_id, fire_time_unix_seconds, task_sequence)    
+);
+
+CREATE TABLE xdb_sys_local_queue(
+    process_execution_id uuid,
+    queue_name VARCHAR(255),
```
longquanzheng, line 72: nit: maybe VARCHAR(31)

persistence/data_models_json_fields.go
```diff
@@ -91,6 +92,119 @@ func (s *StateExecutionSequenceMapsJson) CompleteNewStateExecution(stateId strin
...
+	return json.Marshal(s)
+}
+
+func NewStateExecutionWaitingQueuesFromBytes(bytes []byte) (StateExecutionWaitingQueuesJson, error) {
+	var waitingQueuesJson StateExecutionWaitingQueuesJson
+	err := json.Unmarshal(bytes, &waitingQueuesJson)
+	return waitingQueuesJson, err
+}
+
+func (s *StateExecutionWaitingQueuesJson) Add(stateExecutionId StateExecutionId, command xdbapi.LocalQueueCommand, anyOfCompletion bool) {
```
longquanzheng, line 119: Nit: `Add` => `AddNewLocalQueueCommandForStateExecution`

Question: what if there are multiple local queue commands?

persistence/data_models_enums.go
longquanzheng: Having a new status makes sense. 

Question: is this before or after running status?

persistence/sql/sql_process_store_common.go
```diff
@@ -78,3 +79,69 @@ func insertImmediateTask(
...
+					QueueName: message.GetQueueName(),
+					DedupId:   dedupId,
+					Payload:   message.GetPayload(),
+				},
+			})
+		if err != nil {
+			return err
+		}
+
+		err = tx.InsertImmediateTask(ctx, extensions.ImmediateTaskRowForInsert{
```
longquanzheng, line 132: It would be a lot better if we only create a single task for a batch of messages. If a request sent 10 msgs, they can be processed in a single task to save the processing overhead

persistence/sql/sql_process_store_common.go
```diff
@@ -78,3 +79,69 @@ func insertImmediateTask(
 
 	return tx.InsertImmediateTask(ctx, immediateTaskRow)
 }
+
+func (p sqlProcessStoreImpl) publishToLocalQueue(
+	ctx context.Context, tx extensions.SQLTransaction, processExecutionId uuid.UUID, messages []xdbapi.LocalQueueMessage) error {
+	for _, message := range messages {
+		dedupId := uuid.ParseUUID(message.GetDedupId())
+		if dedupId == nil {
```
longquanzheng, line 87: It's is very rare, but rather than silently swallow the error and use a random UUID, it would be better to just return an error.

persistence/sql/publish_to_local_queue.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	err = p.publishToLocalQueue(ctx, tx, curProcExecRow.ProcessExecutionId, request.Messages)
+	if err != nil {
+		return nil, err
+	}
+
+	return &persistence.PublishToLocalQueueResponse{
+		ProcessExecutionId:  curProcExecRow.ProcessExecutionId,
+		HasNewImmediateTask: len(request.Messages) > 0,
```
longquanzheng, line 70: we probably should just return error before starting an transaction when the len is zero

persistence/sql/process_local_queue.go
```diff
@@ -0,0 +1,137 @@
...
+
+	dedupIdString := request.DedupId.String()
+	assignedStateExecutionId, hasFinishedWaiting := waitingQueues.Consume(xdbapi.LocalQueueMessage{
+		QueueName: request.QueueName,
+		DedupId:   &dedupIdString,
+		Payload:   &request.Payload,
+	})
+
+	// early stop if no state can consume the message
+	if assignedStateExecutionId == nil {
```
longquanzheng, line 70: What if the message can be consumed later? 
A message can be sent before the waitUntil API is completed, or there can be a different async state interested in the message later. 

In additional to what states are interested in what queues, we probably also need to store what are the remaining messages in processExecution for this feature. 

Also another question -- what if the command is waiting for more than one message for the same queue? (maybe we have to move the message from the queue into commandResults)

persistence/sql/process_local_queue.go
```diff
@@ -0,0 +1,137 @@
...
+
+	err = tx.UpdateAsyncStateExecution(ctx, *stateRow)
+	if err != nil {
+		return err
+	}
+
+	// will handle the completion logic in CompleteWaitUntilExecution later.
+
+	// Step 3: delete the task row
+	err = tx.DeleteImmediateTask(ctx, extensions.ImmediateTaskRowDeleteFilter{
```
longquanzheng, line 118: Yeah I think this is a lot easier than using the batch delete. I started to wonder if we just always delete the task during processing. Timer tasks are always doing that. backoff retry is too. It's certainly a lot easier with this model.

persistence/data_models_enums.go
longquanzheng: Figured it out! it's after. So running->waiting->completed

persistence/sql/process_wait_until.go
```diff
@@ -0,0 +1,184 @@
...
+			return err
+		}
+
+		waitingQueues, err := persistence.NewStateExecutionWaitingQueuesFromBytes(prcRow.StateExecutionWaitingQueues)
+		if err != nil {
+			return err
+		}
+
+		for _, localQueueCommand := range localQueueCommands {
+			waitingQueues.Add(request.StateExecutionId, localQueueCommand, request.CommandRequest.GetWaitingType() == xdbapi.ANY_OF_COMPLETION)
```
longquanzheng, line 160: I think it make more sense to add for the whole command request -- the waitUntil API only returns once for those commands. it only have one waitingType(AnyOf or AllOf today)

persistence/sql/process_wait_until.go
```diff
@@ -0,0 +1,184 @@
+// Copyright 2023 XDBLab organization
```
longquanzheng, line 1: nit: if this file is for both ProcessWaitUntil and CompleteWaitUntil, probably rename the file to `wait_until.go` instead

extensions/postgres/schema/xdb_sys_schema.sql
zklgame: Done.

persistence/sql/sql_process_store_common.go
zklgame: Done.

persistence/sql/publish_to_local_queue.go
zklgame: Done.

persistence/sql/process_local_queue.go
zklgame: If not using batch delete, then we also don't need to add a new state status.

persistence/sql/process_wait_until.go
zklgame: Done.

persistence/data_models_json_fields.go
zklgame: There is a loop to add each local queue command:
```
for _, localQueueCommand := range localQueueCommands {
	waitingQueues.AddNewLocalQueueCommandForStateExecution(request.StateExecutionId, localQueueCommand,
		request.CommandRequest.GetWaitingType() == xdbapi.ANY_OF_COMPLETION)
}
```

persistence/sql/sql_process_store_common.go
zklgame: Done.

persistence/sql/process_local_queue.go
zklgame: Oh I missed this part. Now I have added `UnconsumedMessages` and implemented the logic to deal with it.

> what if the command is waiting for more than one message for the same queue?
This logic has been implemented previously in this PR. The related logic is in `func (s *StateExecutionWaitingQueuesJson) Consume(message LocalQueueMessageInfoJson) (*string, bool)`.

persistence/sql/process_wait_until.go
zklgame: Here, the whole local queue command requests have been added into `waitingQueues`. I am not going to support timer request in thie PR.

persistence/sql/process_local_queue.go
```diff
@@ -0,0 +1,178 @@
...
+			return nil, err
+		}
+
+		_, ok := finishedStateExecutionIdToPreviousVersionMap[assignedStateExecutionIdString]
+		if ok {
+			finishedStateExecutionIdToPreviousVersionMap[assignedStateExecutionIdString] = stateRow.PreviousVersion
+		}
+	}
+
+	// Step 3: handle finished wait_until task for finishedStateExecutionIdToPreviousVersionMap
```
zklgame, line 146: Currently, only the local queue commands are considered. If to support other commands, then we need to check the completion type and other command results in this place.

engine/immediate_task_conrrent_processor.go
```diff
@@ -426,3 +425,25 @@ func (w *immediateTaskConcurrentProcessor) composeHttpError(
...
+
+		Messages: task.ImmediateTaskInfo.LocalQueueMessageInfo,
+	})
+	if err != nil {
+		return err
+	}
+
+	// TODO: need a better way to define the NotifyNewImmediateTasks method
+	if resp.HasNewImmediateTask {
+		w.taskNotifier.NotifyNewImmediateTasksNoRequest(task.ShardId)
```
longquanzheng, line 446: `NotifyNewImmediateTasksNoRequest` is probably not needed. Essentially, the `NotifyImmediateTasksRequest` only requires `shardId`. All other fields are optional. I think we can get processExecutionId here. You can ignore other fields if they are not available. (The namespace/processId are  mainly  for future support NoSQL because the partitionKey will be namespace+processId, which is not needed for SQL)

persistence/data_models_json_fields.go
```diff
@@ -139,15 +281,16 @@ type WorkerTaskBackoffInfoJson struct {
 }
 
 type LocalQueueMessageInfoJson struct {
-	QueueName string    `json:"queueName"`
-	DedupId   uuid.UUID `json:"dedupId"`
+	QueueName string               `json:"queueName"`
+	DedupId   uuid.UUID            `json:"dedupId"`
+	Payload   xdbapi.EncodedObject `json:"payload"`
```
longquanzheng, line 286: We probably should remove this payload here. The payload should only live in the local_queue table. Otherwise the process execution will grow too big and cause perf issue to the database. When a state execution succesfully consume the message, we will:
1. select the payload from the table
2. move the payload into the commandResults
3. delete the row from the table.

persistence/data_models_json_fields.go
```diff
@@ -91,6 +92,127 @@ func (s *StateExecutionSequenceMapsJson) CompleteNewStateExecution(stateId strin
 	return nil
 }
 
+type StateExecutionWaitingQueuesJson struct {
+	// { queue_1: {(state_1, state_1_seq): waiting_count, (state_2, state_2_seq): waiting_count, ...} }
```
longquanzheng, line 96: nit: probably just `(state_1, state_1_seq)` ==> `stateExecutionId1`

persistence/data_models_json_fields.go
```diff
@@ -91,6 +92,127 @@ func (s *StateExecutionSequenceMapsJson) CompleteNewStateExecution(stateId strin
...
+
+	if anyOfCompletion {
+		m[stateExecutionIdKey] = 1
+	} else {
+		m[stateExecutionIdKey] += int(count)
+	}
+	s.QueueToStatesMap[command.GetQueueName()] = m
+
+	if anyOfCompletion {
+		s.StateToQueueCountMap[stateExecutionIdKey] = 1
```
longquanzheng, line 144: I think we have some misunderstanding that I may not clearly explained to you earlier). 

The anyOf is on the command level, not on the command messages level. Say we have a commandRequest:
```
{
  WaitingType: AnyOf
  TimerCommands: [
     {TimerCommand1},
     {TimerCommand2}
   ],
   LocalQueueCommands:[
     {
          QueueName: "queue1",
          Count: 2,
     },
     {
          QueueName: "queue2",
          Count: 3,
     },
   ]
}
```
Then it will only complete waiting when any of two timer commands or local queue commands are completed. For the two local queue commands, each will require to get all the messages -- the first one will require to get 2 messages, and the second will require 3.

persistence/data_models_json_fields.go
```diff
@@ -91,6 +92,127 @@ func (s *StateExecutionSequenceMapsJson) CompleteNewStateExecution(stateId strin
...
+		s.UnconsumedMessages = append(s.UnconsumedMessages, message)
+		return nil, false
+	}
+
+	keys := make([]string, 0, len(m))
+	for key := range m {
+		keys = append(keys, key)
+	}
+
+	// E.g., given m as:
```
longquanzheng, line 164: Thanks for having those comments. It is extremely helpful for me to understand. 

I think we are missing something here -- should a LocalQueueCommand(with count>1) consume a message even the message is not enough to fulfill the whole command? 

Consider this example:
stateExeA waits for q1 for 2 msg, and a message comes, so it's assigned to stateExeA but it can't complete(needing one more).

But then stateExeB waits for q1 for 1 msg, should we move the 1 msg to B so it it can complete? 

This is related to what we discussed early for "QueueTypes":
* FCFS type: whoever waits for the queue name will consume the messages for the queue first
* Best first match for command: we won't assign the message unless the whole localQueueCommand can be fulfilled 

Looks this implementation is between the two Lol(very creative btw, I never thought about this idea) -- it's sorting based on current waiting count, and always assign to whoever wait the least. It's like a heuristic optimization. It will works but I feel like it will not be easily understood. 

Based on that thinking, I feel like we probably can change the `QueueToStatesMap` / `StateToQueueCountMap` to just store stateExe to their localQueue commands directly (because merging them with queue names will lose some info, for example, a stateExe could have multiple localQueue commands for the same queueName). And change `UnconsumedMessages` to group by the queue name. When a new messages arrive, just append/add the msg to the UnconsumedMessages and try consume based on the whole localQCommand.

extensions/sql_db_interfaces.go
```diff
@@ -60,13 +60,17 @@ type transactionalCRUD interface {
...
 	UpdateAsyncStateExecution(ctx context.Context, row AsyncStateExecutionRowForUpdate) error
+	UpdateAsyncStateExecutionWithoutCommands(ctx context.Context, row AsyncStateExecutionRowForUpdateWithoutCommands) error
 	BatchUpdateAsyncStateExecutionsToAbortRunning(ctx context.Context, processExecutionId uuid.UUID) error
 	InsertImmediateTask(ctx context.Context, row ImmediateTaskRowForInsert) error
 	InsertTimerTask(ctx context.Context, row TimerTaskRowForInsert) error
 
 	DeleteImmediateTask(ctx context.Context, filter ImmediateTaskRowDeleteFilter) error
 	DeleteTimerTask(ctx context.Context, filter TimerTaskRowDeleteFilter) error
+
+	InsertLocalQueue(ctx context.Context, row LocalQueueRow) error
```
longquanzheng, line 73: Lol, curious where is the deleteLocalQueue

longquanzheng: Actually, to be accurate, maybe we should rename `LocalQueue` to `LocalQueueMessage`

persistence/sql/process_local_queue.go
```diff
@@ -0,0 +1,178 @@
...
+		return nil, err
+	}
+
+	// merge waitingQueues.UnconsumedMessages into request.Messages to consume
+	// waitingQueues.UnconsumedMessages need to be first to ensure the consumption order
+	messages := append(waitingQueues.UnconsumedMessages, request.Messages...)
+	waitingQueues.ClearUnconsumedMessages()
+
+	for _, message := range messages {
+		assignedStateExecutionIdString, hasFinishedWaiting := waitingQueues.Consume(message)
```
longquanzheng, line 71: Nit: it may be more clear to split this into two method:
* ProcessMessage(which put into the unconsumed list)
* TryConsume(see if any messages can be consumed after adding a new message)

persistence/sql/sql_process_store_common.go
```diff
@@ -78,3 +84,89 @@ func insertImmediateTask(
...
+
+		// dealing with user-customized dedupId
+		if message.GetDedupId() != "" {
+			dedupId2, err := uuid.ParseUUID(message.GetDedupId())
+			if err != nil {
+				return false, err
+			}
+
+			// need to check if this message has been published before
+			existingMessage, err := p.session.SelectLocalQueue(ctx, processExecutionId, message.GetQueueName(), dedupId2)
```
longquanzheng, line 107: This checking can be done via the insert query -- it will returns error on conflict

persistence/data_models_json_fields.go
zklgame: If deleting the row from local queue table, how could we check the duplicated dedupId?

longquanzheng:
Ahh great point. We can’t delete it

extensions/postgres/schema/xdb_sys_schema.sql
```diff
@@ -64,4 +64,12 @@ CREATE TABLE xdb_sys_timer_tasks(
...
     info jsonb ,
     PRIMARY KEY (shard_id, fire_time_unix_seconds, task_sequence)    
+);
+
+CREATE TABLE xdb_sys_local_queue(
+    process_execution_id uuid,
+    dedup_id uuid,
+    queue_name VARCHAR(31),
+    payload jsonb,
+    PRIMARY KEY (process_execution_id, dedup_id)
```
zklgame, line 74: @longquanzheng we actually don't need the queue_name in the PKs.

longquanzheng: Very good point! Agreed

persistence/sql/publish_to_local_queue.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	resp, err := p.doPublishToLocalQueueTx(ctx, tx, request)
+
+	if err != nil {
+		err2 := tx.Rollback()
+		if err2 != nil {
+			p.logger.Error("error on rollback transaction", tag.Error(err2))
+		}
+	} else if resp.HasNewImmediateTask { // if there is no NewImmediateTask, it means there is nothing to commit in the transaction
```
zklgame, line 38: @longquanzheng It seems if there is no actual sql operation, commit will fail.

longquanzheng: No operation do you mean because the sql doesn’t have any effect because of duplicates?

zklgame: Yes, if commit, the error message is `failed to commit on failure transaction`.

zklgame: Modified!

engine/immediate_task_conrrent_processor.go
```diff
@@ -426,3 +425,29 @@ func (w *immediateTaskConcurrentProcessor) composeHttpError(
...
+
+		ProcessExecutionId: task.ProcessExecutionId,
+
+		Messages: task.ImmediateTaskInfo.LocalQueueMessageInfo,
+	})
+	if err != nil {
+		return err
+	}
+
+	// TODO: need a better way to define the NotifyNewImmediateTasks method
```
longquanzheng, line 444: I don't quite follow, can you expand more on that?

persistence/data_models.go
```diff
@@ -187,31 +190,71 @@ type (
...
+
+	PublishToLocalQueueResponse struct {
+		ProcessExecutionId  uuid.UUID
+		HasNewImmediateTask bool
+		NotExists           bool
+	}
+
+	ProcessLocalQueueMessagesRequest struct {
+		TaskShardId int32
+		// If TaskSequence is 0, skip deleting the immediate task row, and try to consume unconsumed messages for StateExecutionId
```
longquanzheng, line 242: Took me quite a while to understand this one, still not fully understand yet, but looks like it's for some internal implementation of the `ProcessLocalQueueMessages` persistence API. 

I would recommend to create a different internal request object for this -- put the internal request inside the persistence package so it's not leaked to outside.
```
internalProcessLocalQueueMessagesRequest struct{
       ProcessLocalQueueMessagesRequest
       isNestedRequest bool
       // below are only present when IsNestedRequest is true
       stateExecutionId   *StateExecutionId
       commandWaitingType *xdbapi.CommandWaitingType
}

Or

internalProcessLocalQueueMessagesRequest struct{
       ProcessLocalQueueMessagesRequest
       fromWaitUntilProcessing bool 
       // below are only present when IsNestedRequest is true
       stateExecutionId   *StateExecutionId
       commandWaitingType *xdbapi.CommandWaitingType
}
// seems like this is only when we are consuming when processing the waitUntil?
```
(use lower case to make them internal/private to outside)

persistence/data_models_json_fields.go
```diff
@@ -91,6 +91,164 @@ func (s *StateExecutionSequenceMapsJson) CompleteNewStateExecution(stateId strin
 	return nil
 }
 
+type StateExecutionWaitingQueuesJson struct {
+	// { state_execution_id_1: [ (queue_name_1, count_1), (queue_name_2, count_2), ... ], ... }
+	StateToCommandsMap map[string][]xdbapi.LocalQueueCommand `json:"stateToCommandsMap"`
```
longquanzheng, line 96: nit: `StateToCommandsMap`=> `StateToLocalQueueCommands` or `StateToLocalQueues` for shorter. 

Because I worry that in the future we may want to add something else like global queue, it will be confusing

persistence/data_models_json_fields.go
```diff
@@ -91,6 +91,164 @@ func (s *StateExecutionSequenceMapsJson) CompleteNewStateExecution(stateId strin
...
+	// { queue_name_1: [dedupId_1, dedupId_2, ...], queue_name_2: [dedup_id, ...], ... }
+	UnconsumedMessageQueueDedupIdsMap map[string][]uuid.UUID `json:"unconsumedMap"`
```
longquanzheng, line 98: How about:
```
UnconsumedLocalQueueMessages map[string][]InternalLocalQueueMessage`json:"unconsumedLocalQueueMessages"`

InternalLocalQueueMessage struct{
    DedupId string
    IsFull bool // only false for now until we support including payload
}
     
```
* So that we can support the optimization later
* JSON has pretty bad support for byte array actually. each byte will consume 5? bytes to store, like \xAAAA.

persistence/data_models_json_fields.go
```diff
@@ -91,6 +91,164 @@ func (s *StateExecutionSequenceMapsJson) CompleteNewStateExecution(stateId strin
...
+//
+//	state_1, 1: (q1, 2),
+//	state_1, 2: (q2, 3),
+//	state_3, 1: (q2, 2)
+//
+// If receiving the queue `q2` for twice, then state_3, 1 will consume the two queue `q2`, and StateToCommandsMap becomes:
+//
+//	state_1, 1: (q1, 2),
+//	state_1, 2: (q2, 3),
+func (s *StateExecutionWaitingQueuesJson) Consume(message LocalQueueMessageInfoJson) (*string, []uuid.UUID) {
```
longquanzheng, line 147: nit: just convention, people use "" for empty string rather than pointer.

longquanzheng: Also it would be nice to add a comment about what does the return values mean

persistence/data_models_json_fields.go
```diff
@@ -91,6 +91,164 @@ func (s *StateExecutionSequenceMapsJson) CompleteNewStateExecution(stateId strin
...
+				s.StateToCommandsMap[stateExecutionIdKey][i+1:]...)
+			if len(s.StateToCommandsMap[stateExecutionIdKey]) == 0 {
+				delete(s.StateToCommandsMap, stateExecutionIdKey)
+			}
+
+			return &stateExecutionIdKey, consumedDedupIds
+		}
+	}
+
+	return nil, []uuid.UUID{}
```
longquanzheng, line 174: nit: you can also just return nil for an empty slice

persistence/data_models_json_fields.go
```diff
@@ -91,6 +91,164 @@ func (s *StateExecutionSequenceMapsJson) CompleteNewStateExecution(stateId strin
...
+
+			consumedDedupIds := s.UnconsumedMessageQueueDedupIdsMap[message.QueueName][:int(command.GetCount())]
+
+			s.UnconsumedMessageQueueDedupIdsMap[message.QueueName] =
+				s.UnconsumedMessageQueueDedupIdsMap[message.QueueName][int(command.GetCount()):]
+			if len(s.UnconsumedMessageQueueDedupIdsMap[message.QueueName]) == 0 {
+				delete(s.UnconsumedMessageQueueDedupIdsMap, message.QueueName)
+			}
+
+			s.StateToCommandsMap[stateExecutionIdKey] = append(s.StateToCommandsMap[stateExecutionIdKey][:i],
```
longquanzheng, line 164: Wow, this is so concise. I really like it after understanding it. 
Though it took me a couple minutes to understand:
* it's different from the regular pattern: `slice = append(slice, new eles)`, but just look so closed!!! LOL
* it's not very usual to just update the map/slice when iterating through it(similar in Java, you usually don't update the array when iterating it)

So I would recommend:
* Move this update outside of the for loop. Since it's the last thing to do before return. We could just break here (maybe set a "foundMatched" bool, and do it outside.
* Add a friendly comment/note about this cool operation!

engine/immediate_task_conrrent_processor.go
zklgame: This comment has been removed...

persistence/data_models_json_fields.go
```diff
@@ -91,6 +91,164 @@ func (s *StateExecutionSequenceMapsJson) CompleteNewStateExecution(stateId strin
...
+// (q1, 1), (q2, 2)
+//
+// and isAllOfCompletion as true. Then after ConsumeFor, the UnconsumedMessageQueueCountMap becomes:
+//
+// (q1, 1), (q3, 1)
+//
+// and returns:
+//
+// (true, [q1_dedup_id_1, q2_dedup_id_1, q2_dedup_id_2])
+func (s *StateExecutionWaitingQueuesJson) ConsumeFor(stateExecutionId StateExecutionId, isAllOfCompletion bool) (bool, []uuid.UUID) {
```
longquanzheng, line 194: Nit:
* Maybe `ConsumeFor` --> `CheckCanCompleteLocalQueueWaiting` (it's a bit long, but much clear by just reading the name)
* `isAllOfCompletion`, maybe just use xdbapi.CommandWaitingType enum instead. We will have others in the future.

persistence/data_models_json_fields.go
```diff
@@ -91,6 +91,164 @@ func (s *StateExecutionSequenceMapsJson) CompleteNewStateExecution(stateId strin
...
+			remainingCommands = append(remainingCommands, command)
+			continue
+		}
+
+		consumedDedupIds = append(consumedDedupIds, s.UnconsumedMessageQueueDedupIdsMap[command.GetQueueName()][:int(command.GetCount())]...)
+
+		s.UnconsumedMessageQueueDedupIdsMap[command.GetQueueName()] = s.UnconsumedMessageQueueDedupIdsMap[command.GetQueueName()][int(command.GetCount()):]
+
+		if len(s.UnconsumedMessageQueueDedupIdsMap[command.GetQueueName()]) == 0 {
+			delete(s.UnconsumedMessageQueueDedupIdsMap, command.GetQueueName())
```
longquanzheng, line 217: Still going through the details, but I wonder if we can just keep this method read-only, and the caller will call `CleanupFor` instead for deleting

longquanzheng (commented):
GREAT JOB!
Just haft way though...I will continue tomorrow.

extensions/postgres/transactional.go
```diff
@@ -200,3 +243,27 @@ func (d dbTx) DeleteTimerTask(ctx context.Context, filter extensions.TimerTaskRo
...
+	if err != nil {
+		return err
+	}
+
+	effected, err := result.RowsAffected()
+	if err != nil {
+		return err
+	}
+	if effected != 1 {
+		return conditionalUpdateFailure
```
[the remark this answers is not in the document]
zklgame: Good idea, I modified it to return (insertSuccessfully, error).

persistence/data_models_json_fields.go
zklgame: Good idea.

persistence/data_models_json_fields.go
zklgame: > Also it would be nice to add a comment about what does the return values mean

it's added at the top of the comments.

persistence/data_models_json_fields.go
zklgame: Adding a comment sounds good to me. 

If moving the logic outside, then we need to record 2 break points, one for command, one for state, I feel it's a little more cumbersome to write.

persistence/data_models_json_fields.go
zklgame: This len check for UnconsumedMessageQueueDedupIdsMap needs to be performed everytime a consumption happen, otherwise the data will contain many empty `{}`.

engine/immediate_task_conrrent_processor.go
zklgame: Oh I found it still exists, maybe there were two places with this comment and I only deleted one. Deleted it now.

persistence/data_models.go
zklgame: The meaning of `isNestedRequest ` is not clear to me. There are two cases when we will call `doProcessLocalQueueMessageTx`:

- in the client.publishToLocalQueue: do normal consume
- when a new wait_until task was executed and need to check whether it should consume the unconsumed messages: do CheckCanCompleteLocalQueueWaiting.

After either step above, the method will update state execution, update process execution, complete possible wait_until statuses.

And the TaskSequenceId is used to distingwish the two cases.

zklgame: Acutally, doing another nest level on ProcessLocalQueueMessagesRequest looks a little more cumbersome to me.

persistence/data_models_json_fields.go
longquanzheng: You can label the first for loop like this https://stackoverflow.com/questions/51996175/how-to-break-out-of-nested-loops-in-go

persistence/data_models_json_fields.go
longquanzheng: Okay, I think I see what you meant. 

I think then it should be `ConsumeWithCheckingLocalQueueWaitingComplete`, instead of `CheckCanCompleteLocalQueueWaiting` (sorry for the back and forth). 

But I still a little confused why we have this and the `Consume` method above.

persistence/data_models.go
longquanzheng: > And the TaskSequenceId is used to distingwish the two cases.
Overloading this field is not necessary and caused confusion. 


It should be very small refactoring on what you have done:
```golang
func (p ...)ProcessNewLocalQueueMessagesTask(req ...){
    internalReq := {
        ProcessLocalQueueMessagesRequest: req
        fromWaitUntilProcessing: false
    }
    ...
    resp, err := p.doProcessNewLocalQueueMessagesTaskTx(tx, internalReq)
    ...
}

func (p ...)doProcessWaitUntilExecutionTx(tx, req ...){
   ...
   internalReq := {
       ProcessLocalQueueMessagesRequest: {...}
       fromWaitUntilProcessing: true,
       stateExecutionId: ...
       commandWaitingType: ...
   }
    resp, err := p.doProcessNewLocalQueueMessagesTaskTx(tx, internalReq)
}
```

> in the client.publishToLocalQueue: do normal consume
btw I believe you meant processing the immediate tasks in async service, because we only process in the background.

longquanzheng (commented):
Finally read through. Overall, really GREAT JOB. However, I think we still need to improve on this feature quite a bit:

I just realized that we have to introduce a `commandId` like in iWF: https://github.com/xdblab/xdb-apis/blob/c2b066c23943d17c6a0a673e18298e8eb5d30edb/api-schema/xdb.yaml#L319
A commandId can uniquely identify a command within a commandRequest. (it's also necessary for future support for anyCombinationOf waiting type, because the combination is specified by commandIDs)

Otherwise, it seems problematic to sync the message consumption from procExe to stateExe. 

Some notes:
* We don't have to require the commandID to be set by user, it can be set automatically by SDK or server
* If set by user (return from waitUntil response), we have to check the duplication
* It is probably better to use integer than string for commandID. It will be easier for us to generate automatically, and more efficient to store and process later 

OpenAPI doesn't allow using map directly, so we have to create a new struct:
```golang
InternalCommandRequest struct{
    WaitingType: enum
    commands: map[commandID] InternalCommand
}

InternalCommand {
  CommandType: timer/localqueue
  TimerCommand: *xdbapi.TimerCommand //only applicable for timer commandType
  LocalQueueCommand: *xdbapi.LocalQueueCommand //...
}
```

Also, I think it's not necessary to share too many consuming logic between processLocalQTask and processWaitUntil. They are too different by nature -- processLocalQTask will process for all the states, while processWaitUntil only care about the state execution itself ( and it's technically an optimization to avoid creating an extra immediate task when the messages are enough to complete waiting immediately). 

I am thinking that the structure will be like this to be more efficient(and hopefully clear as well):

```golang

// I feel like LocalQueue is a better name(for future extension for global queu). it just includes the unconsumed messages and current waiting commands like you are doing
LocalQueues struct{
   waitingStates: map[stateExe]map[commandId]InternalCommand
   unconsumeMsgs : map[queueName] []InternalMsg
} 

func (q *LocalQueues) AddStateExecution( statExeId, commands) { 
 // add all commands at once
}

func (q *LocalQueues) RemoveStateExecution( statExeId) { 
 // keep the remove in this struct to be more clear
}

func (q *LocalQueues) AddMessageAndConsumeForExistingCommands( msg ) (
stateExeId, commandId, []msg) { 
// add a new message and try consume the existing commands 
// also return commandId so that it can be used for syncing with stateExe later
}

func (q *LocalQueues)  ConsumeForNewCommand( newCmd ) ([]msg) { 
// separate method for consuming new command, without adding a new msg
}

func (p *ProcessStore) doProcessNewLocalQueueMessages( req ){
     //1. select the proc exec for update, same as you did
     
     // 2. consume for all, mostly the same as you did
     for loop req.Msgs {
       stateExeId, cmdId, msg := AddMessageAndConsumeForExistingCommands(..)
       assignedStateExecutionIdToMessagesMap[stateExe] = append(...)
     }
     
     3. for loop assignedStateExecutionIdToMessagesMap {
         3.1 select stateExecution for update 
         3.2 select all msg payload from local_queue_msg
         3.3 update the stateExe row:
              3.3.1: add/append the completed command to commandResults with msgs 
              3.3.2: check the commandResults with commandRequest to see if it's done. If done, then 
              3.3.2.1 call RemoveStateExecution to remove the stateExe from the LocalQueues of procExe
              3.3.2.2 update the stateExe status to executeRunning 
              3.3.2.3 setHasNewImmediateTask
     }
     
     4. InsertNewImmediateTask if hasNewImmediateTask 
     5. Delete current task
     6. Finally update the procExe 
}

func (p *ProcessStore) doProcessWaitUntil(...){
     1. for empty command, complete and return like before

     2. select the prcExe for update first
     
     3. immeidately try consume existing messages if there is any (len(unconsumedMsg) > 0)
     var immediateCompletedCmds map[cmdId][]InternalMsg
     forloop commandRequest.localQueueCommand {
          call ConsumeForNewCommand, if succeeded, append to immediateCompletedCmds
     }
     
     4. if len(immediateCompletedCmds)>0
          4.1 select from local_queue_msg      
          4.2 . set commandResults 
          
      5. check if the commandRequest is able to complete, 
      5.1 if yes, set the status to executeRunning and set hasNewImmediateTask
      5.2 if no, add the remaining command to procExe.localQueues

     6. update procExe, update stateExe
     7. process new publish msg to local_que_messages
     8. insert new immediate task if needed
}
```

longquanzheng (approved):

persistence/data_models.go
zklgame: Need to think more about how to refactor. Because I feel the ProcessLocalQueueMessagesRequest is even not necessary in the internalReq sent by `doProcessWaitUntilExecutionTx` (we can move the few used fields out from the request).

longquanzheng: Yeah you actually can ignore this because my latest comment is to not share this method between the two persistence APIs https://github.com/xdblab/xdb/pull/63#pullrequestreview-1697935884

we can sync on zoom for this too

persistence/data_models_json_fields.go
zklgame: > But I still a little confused why we have this and the Consume method above.

To maintain the map as `{}` but not `{{}, {}, {}}`?

longquanzheng: Yeah I got it now. I probably should remove this comment

longquanzheng: When reviewing code, I used a review session to keep accumulate comments  and publish all at the end. Some of the earlier comments became stale that I forgot to remove before publishing

persistence/data_models_json_fields.go
```diff
@@ -91,6 +91,174 @@ func (s *StateExecutionSequenceMapsJson) CompleteNewStateExecution(stateId strin
...
+//
+//	state_1, 1: (q1, 2),
+//	state_1, 2: (q2, 3),
+//	state_3, 1: (q2, 2)
+//
+// If receiving the queue `q2` for twice, then state_3, 1 will consume the two queue `q2`, and StateToLocalQueueCommandsMap becomes:
+//
+//	state_1, 1: (q1, 2),
+//	state_1, 2: (q2, 3),
+func (s *StateExecutionWaitingQueuesJson) Consume(message LocalQueueMessageInfoJson) (string, []InternalLocalQueueMessage) {
```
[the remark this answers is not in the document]
longquanzheng: This is probably stale after this final comment https://github.com/xdblab/xdb/pull/63#pullrequestreview-1697935884

persistence/sql/wait_until.go
```diff
@@ -0,0 +1,213 @@
...
+		StateId:            request.StateId,
+		StateIdSequence:    request.StateIdSequence,
+	})
+}
+
+func (p sqlProcessStoreImpl) updateWaitUntilExecution(
+	ctx context.Context, tx extensions.SQLTransaction, request persistence.ProcessWaitUntilExecutionRequest,
+) (*persistence.ProcessWaitUntilExecutionResponse, error) {
+	// to consume unconsumed messages when this state has localQueueCommands AND there are unconsumed messages in the waitingQueue
+	toConsumeUnconsumedMessages := false
```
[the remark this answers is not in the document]
longquanzheng: This is probably stale after this final comment https://github.com/xdblab/xdb/pull/63#pullrequestreview-1697935884

persistence/sql/wait_until.go
```diff
@@ -0,0 +1,213 @@
...
+		}
+
+		toConsumeUnconsumedMessages = len(waitingQueues.UnconsumedLocalQueueMessages) > 0
+
+		prcRow.StateExecutionWaitingQueues, err = waitingQueues.ToBytes()
+		if err != nil {
+			return nil, err
+		}
+
+		err = tx.UpdateProcessExecution(ctx, *prcRow)
```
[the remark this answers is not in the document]
zklgame: Here, UpdateProcessExecution will only perform once for all the commands.

But later this method calls doProcessLocalQueueMessageTx, I might refactor on it to reduce the second select and update.

PR #68, opened 2023-10-27
Reduce sql operations

zklgame:

Improve the logic of handling local queue messages and make exactly-once update to each table row.

## What has changed

- API change: ...
- Golang SDK change: ...
- Server Component 1: ...
- Server Component 2: ...

## Checklist before merge


zklgame:
This refactor uses a single `doProcessLocalQueueCommandsAndMessagesTx` method to handle the logic for both updateWaitUntil and processLocalQueueMessages. Compared to the proposal in https://github.com/xdblab/xdb/pull/63#pullrequestreview-1697935884, the pros and cons of this refactor are:

- pros:  highly resuse the codes and reduce maintainance effort for related logic.
- cons: might not as straightforward as the other one and sacrifice some readability.

In terms of functionality, these two methods can achieve the same results and performance.

@longquanzheng What's your preference now?

longquanzheng:
> This refactor uses a single `doProcessLocalQueueCommandsAndMessagesTx` method to handle the logic for both updateWaitUntil and processLocalQueueMessages. Compared to the proposal in [#63 (review)](https://github.com/xdblab/xdb/pull/63#pullrequestreview-1697935884), the pros and cons of this refactor are:
> 
> * pros:  highly resuse the codes and reduce maintainance effort for related logic.
> * cons: might not as straightforward as the other one and sacrifice some readability.
> 
> In terms of functionality, these two methods can achieve the same results and performance because they both update the related rows in tables only once.
> 
> @longquanzheng What's your preference now?

Yeah,  it's indeed a better/improved version than your previous one, and technically the same perf as what I proposed. I can clearly see the more code reuse than what I proposed:
* reuse the `SelectProcessExecutionForUpdate` / `UpdateProcessExecution`
* reuse the `SelectAsyncStateExecutionForUpdate` / `UpdateAsyncStateExecution`
* reuse assignedStateExecutionIdToMessagesMap, etc

And as you pointed out the readability:
* There are three times that we have to check `request.ProcessLocalQueueCommandsAndTryConsumeRequest != nil` to branch which actually brings quite some complexity to understand
* The nested structure of `ProcessLocalQueueCommandsAndMessagesResponse` makes the code long and tedious to read
* This reuse at doAbcTx is a new pattern in this code base, unless it's a great fit, adding a new coding pattern will bring down the consistency of the code base. It's pretty hard to say it's great fit or not. I can see process messages from task is also processing existing command of all state Executions, share some similarity with processing new command from waitUntil 😮‍💨 

Finally for tradeoff in my own opinion, these additional code reuse with this is not as valuable as readability and **extensibility**.
* Reusing the `SelectProcessExecutionForUpdate` / `UpdateProcessExecution` /`SelectAsyncStateExecutionForUpdate` / `UpdateAsyncStateExecution` will limit how we can extend the waitUntil later (very soon, we will add timer command, and future GlobalQueue command). Or we will have to make this persistence API a giant combo for the new commands. 
* It would be in-efficient to share the same  `SelectProcessExecutionForUpdate` / `UpdateProcessExecution` /`SelectAsyncStateExecutionForUpdate` / `UpdateAsyncStateExecution` -- we may update more fields in WaitUntil processing, but just very few fields in processNewMessagesTask. We may want to use separate SQL API to improve the perf of write overhead (the JSON blob will have high overhead)

persistence/sql/sql_process_store_common.go
```diff
@@ -158,3 +158,103 @@ func (p sqlProcessStoreImpl) publishToLocalQueue(
...
+		for _, localQueueCommand := range commandRequest.LocalQueueCommands {
+			if localQueueToMessageCountMap[localQueueCommand.QueueName] < int(localQueueCommand.GetCount()) {
+				return false
+			}
+		}
+		return true
+	case xdbapi.EMPTY_COMMAND:
+		return true
+	default:
+		return true
```
longquanzheng, line 240: we should return an error for this, or even just panic is fine as this is a major code bug

persistence/sql/sql_process_store_common.go
```diff
@@ -158,3 +158,103 @@ func (p sqlProcessStoreImpl) publishToLocalQueue(
...
+	}
+}
+
+func (p sqlProcessStoreImpl) completeWaitUntilWaiting(ctx context.Context, tx extensions.SQLTransaction,
```
longquanzheng, line 244: nit: I don't feel like the three operations are worth putting into this method name `completeWaitUntilWaiting` as the name doesn't seem easily guess with what it is doing, until jumping into the code. I tried to find a better name but still, the three operations (clean up queues, set status and insert immediate tasks are too different). I would rather keep copy paste for now. to make ti more clear to read the code. 

The best name I can think of is (miscOperationsForCompleteWaitUntil 🤦 ) "completeWaitUntilWaiting" makes it like updating the state execution

persistence/sql/wait_until.go
```diff
@@ -84,7 +84,7 @@ func (p sqlProcessStoreImpl) doProcessWaitUntilExecutionTx(
 	}, nil
 }
 
-func (p sqlProcessStoreImpl) CompleteWaitUntilExecution(
+func (p sqlProcessStoreImpl) completeWaitUntilExecution(
```
longquanzheng, line 87: 👍

longquanzheng (approved):
just some minor comments on naming. 

Also consider:
waitingQueues --> localQueues because
* it is just for local queues 
* it stores both the waiting info of state execution for local queues and the unconsumed messages of the local queues
* waiting queue doesn't seem fluent/accurate

persistence/sql/sql_process_store_common.go
zklgame: True, this is another too-early-optimization lol. Let's keep copy and paste for now.

PR #79, opened 2023-11-06
add support for process timeout

duoertai:

## What has changed

- API change: ...
- Golang SDK change: ...
- Server Component 1: ...
- Server Component 2: ...

## Checklist before merge


engine/timer_task_conrrent_processor.go
```diff
@@ -113,14 +114,35 @@ func (w *timerTaskConcurrentProcessor) processTimerTask(
...
+	task persistence.TimerTask,
+) error {
+	resp, err := w.store.ProcessTimerTaskForProcessTimeout(w.rootCtx, persistence.ProcessTimerTaskRequest{
+		Task: task,
+	})
+	if err != nil {
+		return err
+	}
+
+	if resp == nil {
```
longquanzheng, line 135: In Golang, usually we don't check this. `err != nil` has indicated that this field is not nil, otherwise the next line will panic as well as the same results.

service/api/service_impl.go
```diff
@@ -43,9 +43,17 @@ func NewServiceImpl(cfg config.Config, store persistence.ProcessStore, logger lo
 func (s serviceImpl) StartProcess(
 	ctx context.Context, request xdbapi.ProcessExecutionStartRequest,
 ) (response *xdbapi.ProcessExecutionStartResponse, retErr *ErrorWithStatus) {
+	timeoutUnixSeconds := 10
```
longquanzheng, line 46: I don't think we want to have a default timeout like that. 

Having unlimited timeout is an important feature for uses. If there is no timeout specified, it should be infinite timeout. (same as in Temporal/iWF, Cadence actually require timeout to be set on start API)

longquanzheng (commented):
Great Job overall!!

lets make timeout optional for process

persistence/data_models.go
```diff
@@ -72,6 +75,28 @@ type (
 		ShardId                      int32
 	}
 
+	GetProcessExecutionByIdRequest struct {
```
longquanzheng, line 78: Looks like unused code

duoertai:
in backoff test failure case, the worker backoff timer is never fired when it happens, although the process timeout timer task will always fire
the failure doesn't always happen though

persistence/sql/convert_timer_task_to_immediate_task.go
```diff
@@ -35,7 +35,7 @@ func (p sqlProcessStoreImpl) ConvertTimerTaskToImmediateTask(
 
 	return &persistence.ProcessTimerTaskResponse{
 		HasNewImmediateTask: true,
-	}, nil
+	}, err
```
longquanzheng, line 38: 👍

engine/timer_task_queue.go
```diff
@@ -271,8 +272,8 @@ func (w *timerTaskQueueImpl) triggeredPolling() {
 			w.currMaxLoadedTaskSequence = resp.MaxSequenceInclusive
 
 			// add the new tasks into the heap
-			for _, task := range resp.Tasks {
```
longquanzheng, line 274: 👍

persistence/sql/process_timer_task_for_timer_command.go
```diff
@@ -33,7 +33,7 @@ func (p sqlProcessStoreImpl) ProcessTimerTaskForTimerCommand(
 		}
 	}
 
-	return resp, nil
```
longquanzheng, line 36: 👍  very good catch.

longquanzheng (approved):
great job!!

PR #97, opened 2023-11-14
Add RPC

zklgame:

## What has changed

- API change link: ...
- Golang SDK change link: ...
- Server Component 1: ...
- Server Component 2: ...

[It's recommended to write integration test in Golang SDK repo, and enabled in this server repo first, 
without enabling in the SDK repo. After this PR is merged, enable and merge the integration test in the SDK repo]

[Alternatively if Java/other SDK repo is preferred, then just test locally against server PR. 
After the server PR is merged, merge the integration test in the SDK repo]

## Checklist before merge


service/api/service_impl.go
```diff
@@ -149,6 +151,111 @@ func (s serviceImpl) PublishToLocalQueue(
...
+) (response *xdbapi.ProcessExecutionRpcResponse, retErr *ErrorWithStatus) {
+	latestPrcExe, err := s.store.GetLatestProcessExecution(ctx, data_models.GetLatestProcessExecutionRequest{
+		Namespace: request.GetNamespace(),
+		ProcessId: request.GetProcessId(),
+	})
+	if err != nil {
+		return nil, s.handleUnknownError(err)
+	}
+
+	if latestPrcExe.NotExists || latestPrcExe.Status != data_models.ProcessExecutionStatusRunning {
```
longquanzheng, line 165: in IWF, a read-only RPC is allowed for a closed workflow execution. Do you think we can still support it?

service/api/service_impl.go
```diff
@@ -149,6 +151,111 @@ func (s serviceImpl) PublishToLocalQueue(
...
+		TaskShardId: persistence.DefaultShardId,
+	})
+	if err != nil {
+		return nil, s.handleUnknownError(err)
+	}
+	if updateResp.FailAtUpdatingGlobalAttributes {
+		s.logger.Warn("failed to update global attributes")
+		return nil, NewErrorWithStatus(
+			http.StatusFailedDependency,
+			"Failed to write global attributes, please check the error message for details: "+updateResp.UpdatingGlobalAttributesError.Error())
```
longquanzheng, line 240: Error handling is critical for building reliable applications. I think we need to return the error code to applications as mentioned in https://github.com/xdblab/xdb-apis/pull/24#discussion_r1393486766

persistence/sql/update_process_execution_from_rpc.go
```diff
@@ -0,0 +1,121 @@
...
+	// Step 2: handle state decision
+
+	sequenceMaps, err := data_models.NewStateExecutionSequenceMapsFromBytes(prcRow.StateExecutionSequenceMaps)
+	if err != nil {
+		return nil, err
+	}
+
+	// if getting next states and there are running state(s), return error
+	if len(request.StateDecision.GetNextStates()) > 0 && len(sequenceMaps.PendingExecutionMap) > 0 {
+		return nil, fmt.Errorf("cannot start new states when there are running state(s)")
```
longquanzheng, line 71: I don't think this is a valid assumption. There are use cases that need to start new thread while some states are running

service/api/service_impl.go
```diff
@@ -149,6 +151,111 @@ func (s serviceImpl) PublishToLocalQueue(
 	return nil
 }
 
+func (s serviceImpl) Rpc(
+	ctx context.Context, request xdbapi.ProcessExecutionRpcRequest,
+) (response *xdbapi.ProcessExecutionRpcResponse, retErr *ErrorWithStatus) {
+	latestPrcExe, err := s.store.GetLatestProcessExecution(ctx, data_models.GetLatestProcessExecutionRequest{
```
longquanzheng, line 157: I don't think it's safe outside a transaction. The process execution could get terminated or even deleted right after this query.

service/api/service_impl.go
zklgame: http.StatusFailedDependency means 424, this is the code.

service/api/service_impl.go
zklgame: And you cannot put it into transaction because later there will be a API call to the worker. This is similar to how the waitUntil/Execute API works.

config/config.go
```diff
@@ -54,6 +54,8 @@ type (
 		InternalHttpServer HttpServerConfig `yaml:"internalHttpServer"`
 		// ClientAddress is the address for API service to call AsyncService's internal API
 		ClientAddress string `yaml:"clientAddress"`
+		// Rpc is the config for rpc calls
+		Rpc RpcConfig `yaml:"rpc"`
```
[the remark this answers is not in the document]
zklgame: True, modified.

service/api/service_impl.go
zklgame: Sounds good. I moved the second checking condition later to skip writing if the process is not running after receiving the RPC response.

persistence/sql/update_process_execution_from_rpc.go
zklgame: Removed it.

service/api/service_impl.go
longquanzheng: Why we need this error code -- 
the user code will make different decisions on this. If column value is conflict, or database schema is invalid, they should return different error code and different UI message in the application.

zklgame: You wrote it:
https://github.com/xdblab/xdb/blob/689585f839c05e29ffc0b3864e95fd002bd34b06/persistence/sql/complete_execute.go#L49

longquanzheng: This is in async state, we don’t have a good way to return error. We can discuss this in zoom

longquanzheng: And I also need to improve the error code in start process too to return error code

service/api/service_impl.go
zklgame: Then we need to re-design ALL the transactions that query the table `xdb_sys_latest_process_executions` to lock the row in `xdb_sys_latest_process_executions` as the first step in each transation.

service/api/service_impl.go
zklgame: > I meant the error code from the database driver.

The error code from the database driver should be returned by the `updateGlobalAttributesIfNeeded` method first. And here in the implementation for RPC, the `UpdatingGlobalAttributesError` is assigned to whatever returned from `updateGlobalAttributesIfNeeded`, and it's returned to SDK with `NewErrorWithStatus`.

So if you wanna imporve on the error message. you can just update `updateGlobalAttributesIfNeeded` itself in another PR. Currently, I will just return whatever the err is from the `updateGlobalAttributesIfNeeded` to SDK.

service/api/service_impl.go
zklgame: There will be two places to GetLatestProcessExecution. One is at the beginning of the RPC method, by session, to get the process info to call worker API. Another is after getting the worker response, use tx to get the latest process execution to lock.

zklgame: The GetLatestProcessExecution before worke RPC call will still be in the session. Once getting the worker RPC response, we will directly use the `tx.SelectProcessExecutionForUpdate` to lock.

zklgame: And two other places in doStopProcessTx and doPublishToLocalQueueTx can use sessions to SelectLatestProcessExecution, I don't think there will be race condition. Can you just provide a detailed example to prove the opposite side?

persistence/sql/common.go
```diff
@@ -60,3 +63,100 @@ func createGetTimerTaskResponse(
...
+				return nil, err
+			}
+		}
+	}
+
+	// If the process was previously configured to gracefully complete and there are no states running,
+	// then gracefully complete the process regardless of the thread close type set in this state.
+	// Otherwise, handle the thread close type set in this state.
+
+	toGracefullyComplete := procExecWaitToComplete && len(sequenceMaps.PendingExecutionMap) == 0
```
longquanzheng, line 117: just naming nit, maybe:

```
shouldGracefulComplete := procExecGracefulCompleteRequested && len(sequenceMaps.PendingExecutionMap) == 0
```
related to the issue: https://github.com/xcherryio/xcherry/issues/66

`GracefulCompleteRequested` should make it more clear that there was a gracefulComplete decision was made before.

persistence/sql/update_process_execution_from_rpc.go
```diff
@@ -0,0 +1,121 @@
...
+) (*data_models.UpdateProcessExecutionFromRpcResponse, error) {
+	hasNewImmediateTask := false
+
+	// lock process execution row first
+	prcRow, err := tx.SelectProcessExecutionForUpdate(ctx, request.ProcessExecutionId)
+	if err != nil {
+		return nil, err
+	}
+
+	// skip the writing operations on a closed process
```
longquanzheng, line 49: Instead of silently skip this update, it's better to return an error(maybe conflict error) for this case. 

users should be informed that the update is rejected otherwise they won't have the correct expectation for what has happened. (this is the behavior in iWF today without any extra code because Cadence/Temporal will throw error when sending signals to a closed workflow execution)

service/api/service_impl.go
```diff
@@ -149,6 +151,117 @@ func (s serviceImpl) PublishToLocalQueue(
...
+			Input:                  request.Input,
+			LoadedGlobalAttributes: &loadGlobalAttributeResponse,
+		},
+	).Execute()
+	if httpResp != nil {
+		defer httpResp.Body.Close()
+	}
+
+	if utils.CheckHttpResponseAndError(err, httpResp, s.logger) {
+		return nil, NewErrorWithStatus(
```
longquanzheng, line 213: There could be more useful info that the caller need to get from the worker execution. 

In the API, we have decided a workerErrorResponse, and we should try to extract it from the http body to caller. 

See similar in iWF server: https://github.com/indeedeng/iwf/blob/main/service/common/rpc/invoke.go#L59

The reason is that worker may want to return some useful error type for caller to make different decision on.

longquanzheng (commented):
great job (esp the refactoring)

persistence/data_models/data_models.go
```diff
@@ -277,6 +296,54 @@ type (
...
+		TaskShardId int32
+	}
+
+	UpdateProcessExecutionFromRpcResponse struct {
+		HasNewImmediateTask            bool
+		FailAtUpdatingGlobalAttributes bool
+		UpdatingGlobalAttributesError  error
+	}
+
+	HandleStateDecisionRequest struct {
```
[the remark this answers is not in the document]
zklgame: Will move them into sql/common.go.

persistence/data_models/data_models.go
```diff
@@ -47,12 +47,31 @@ type (
 		NotExists bool
 	}
```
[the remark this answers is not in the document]
zklgame: Will move them into sql/common.go.

persistence/sql/common.go
zklgame: I think it will be better to make another PR for the issue, rather than partially renaming something in this PR to make naming inconsistent.

persistence/sql/update_process_execution_from_rpc.go
zklgame: Will return StatusNotFound

service/api/service_impl.go
zklgame: Let's make another PR to handle it. Several places need to modify accordingly, not only here. Created https://github.com/xcherryio/xcherry/issues/98.

longquanzheng (approved):

PR #122, opened 2024-03-15
add cluster support

zklgame:

Add cluster option for the Async service.

```
go run cmd/server/main.go --config ./config/development-postgres-cluster.yaml
```

## TODO in next MRs

- Recovery for failed Async server.
  - when there is any member change, use delegates to call a new API in each async server to add/remove queues
- remove all the defaultShard and store tasks of the same execution in the same shard.


config/config.go
```diff
@@ -67,6 +72,12 @@ type (
 		// See net.Dial for details of the address format.
 		// For more details, see https://blog.cloudflare.com/the-complete-guide-to-golang-net-http-timeouts/
 		Address string `yaml:"address"`
+		// Used in AsyncServiceConfig with AsyncServiceModeConsistentHashingCluster mode only.
+		// Multiple Address seperated by comma.
+		ClusterAddresses string `yaml:"clusterAddresses"`
+		// Used in AsyncServiceConfig with AsyncServiceModeConsistentHashingCluster mode only.
+		// These are addresses used by memberlist.
+		ClusterAdvertiseAddresses string `yaml:"clusterAdvertiseAddresses"`
```
longquanzheng, line 80: I would suggest to put them under `AsyncServiceConfig.ClusterConfig`

longquanzheng: Also let's add more details on what are those configs for and how to use them

service/async/default_server.go
```diff
@@ -28,16 +30,71 @@ type defaultSever struct {
...
 	cfg config.Config,
 	processStore persistence.ProcessStore,
 	visibilityStore persistence.VisibilityStore,
 	logger log.Logger,
-) Server {
+) []Server {
+	var servers []Server
+	addressToServerMap := map[string]Server{}
+
+	serverAddresses := strings.Split(cfg.AsyncService.ClientAddress, ",")
```
longquanzheng, line 43: `cfg.AsyncService.ClientAddress` is the address for api service to call async service. API service needs to notify the async serivce's specific node/shard. But api service doesn't need to use memberlist to lookup. Instead, api service will send to a random node, and then the receiving async service node may forward the request to the right node of the shard.

Because api service doesn't need to know which node own the shard, it can send request behind a LBS.

longquanzheng (commented):
As a hierarchy of the  cluster mode, it will be nice to:
* Each async service process is a standard deployment in K8s(Marvin) so the number of port needs to be static.
* Each async service process is composed by several shards, each shard is composed by immediateQueue + delayQueue.

cluster/delegate.go
```diff
@@ -0,0 +1,55 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package cluster
+
+import (
+	"encoding/json"
+	"log"
+)
+
+type ClusterDelegate struct {
```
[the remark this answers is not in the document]
zklgame: It's used to save the meta data. We use meta data to get the target client address of a node.

service/async/default_server.go
zklgame: The current way is:
1. api server send the internal request to a random async server
2. the random async server forward the request to the target async server 

This is the same as what you described above. See the codes near the comment `// randomly send the request to an async service`.

PR #123, opened 2024-03-19
Split configs for cluster + cluster membership

zklgame:
This MR did:
- Seperate the config for creating api and async servers respectively.
- Add cluster membership mangement, including re-balancing.

TODO in other MRs:
- remove all the default shard id


engine/immediate_task_queue.go
```diff
@@ -68,14 +68,21 @@ func NewImmediateTaskQueueImpl(
 
 func (w *immediateTaskQueueImpl) Stop(ctx context.Context) error {
 	// close timer to prevent goroutine leakage
-	w.pollTimer.Close()
```
zklgame, line 71: I found timer.close() will cause error, need to use stop(). Though don't know why.

config/config.go
```diff
@@ -43,6 +44,10 @@ type (
 		HttpServer HttpServerConfig `yaml:"httpServer"`
 		// Rpc is the config for rpc calls
 		Rpc RpcConfig `yaml:"rpc"`
+		// AsyncAddresses are the addresses for API service to call AsyncServices' internal APIs
+		AsyncAddresses []string `yaml:"asyncAddresses"`
+		// the total shard count in AsyncServices. default to be 1.
+		AsyncShard int `yaml:"asyncShard"`
```
longquanzheng, line 50: It's better to use the same value below to avoid mistakes(what if they are adifferent) and I would prefer to put them together

config/config.go
```diff
@@ -86,6 +91,15 @@ type (
 		MaxHeaderBytes    int           `yaml:"maxHeaderBytes"`
 	}
 
+	MembershipConfig struct {
+		// total shard count across all the async servers
+		Shard int `yaml:"shard"`
```
longquanzheng, line 96: `numShards` or `shards`

config/config.go
```diff
@@ -55,8 +60,8 @@ type (
 		// InternalHttpServer is the config for starting a http.Server
```
longquanzheng, line 60: maybe rename and/or add some comments to explain this vs `AsyncServiceAddresses`. This is for binding, where `AsyncServiceAddresses` is for resolving. 

Maybe this is `InternalAddressToBind`

service/async/service_impl.go
```diff
@@ -13,19 +13,24 @@ import (
...
+	"time"
 )
 
 type asyncService struct {
 	rootCtx context.Context
 
 	taskNotifier engine.TaskNotifier
 
-	immediateTaskQueue     engine.ImmediateTaskQueue
+	immediateTaskQueueMap  map[int32]engine.ImmediateTaskQueue
```
longquanzheng, line 26: maybe `immediateTaskQueuePerShard`

engine/interfaces.go
```diff
@@ -31,6 +33,7 @@ type ImmediateTaskQueue interface {
 	// TriggerPollingTasks exposes an API to be called by TaskNotifier
 	TriggerPollingTasks(request xcapi.NotifyImmediateTasksRequest)
 	Stop(ctx context.Context) error
+	UpdateShardId(shardId int32)
```
longquanzheng, line 36: I don't follow this, why we need to update shardId? We can sync offline

config/config.go
```diff
@@ -86,6 +91,15 @@ type (
 		MaxHeaderBytes    int           `yaml:"maxHeaderBytes"`
 	}
 
+	MembershipConfig struct {
+		// total shard count across all the async servers
+		Shard int `yaml:"shard"`
+		// the advertise address to bind
```
longquanzheng, line 97: It's better to have a different `BindAddress` and default to `AdvertiseAddress`. 

In production env this could be different for networking setup(security etc). User may just bind for 0.0.0.0:1234 for binding but there is a network router to use node1:1234 as the real address

service/async/interfaces.go
```diff
@@ -18,5 +18,13 @@ type Service interface {
...
 	NotifyPollingTimerTask(req xcapi.NotifyTimerTasksRequest) error
+	NotifyRemoteImmediateTaskAsyncInCluster(req xcapi.NotifyImmediateTasksRequest, serverAddress string)
+	NotifyRemoteTimerTaskAsyncInCluster(req xcapi.NotifyTimerTasksRequest, serverAddress string)
 	Stop(ctx context.Context) error
+	ReBalance(assignedShardIds []int32)
+}
+
+type Membership interface {
+	GetServerAddress() string
+	GetServerAddressFor(shardId int32) string
```
longquanzheng, line 29: `GetCurrentAdvertiseAddress` and `GetAdvertiserAddressForShard`

config/config.go
```diff
@@ -55,8 +60,8 @@ type (
 		// InternalHttpServer is the config for starting a http.Server
 		// to serve some internal APIs
 		InternalHttpServer HttpServerConfig `yaml:"internalHttpServer"`
-		// ClientAddress is the address for API service to call AsyncService's internal API
-		ClientAddress string `yaml:"clientAddress"`
+		// Membership is the config for cluster members
```
longquanzheng, line 63: Maybe `ClusterConfig` so that it can be more generic to include `numShards` and more in the future

engine/timer_task_queue.go
```diff
@@ -91,20 +91,27 @@ func NewTimerTaskQueueImpl(
...
 	if req.ShardId != w.shardId {
-		panic("shardId doesn't match")
+		w.logger.Error(fmt.Sprintf("shardId doesn't match: %d - %d", req.ShardId, w.shardId))
+		return
 	}
 	w.triggeredPollingChan <- req
 }
 
+func (w *timerTaskQueueImpl) UpdateShardId(shardId int32) {
+	w.shardId = shardId
```
longquanzheng, line 112: I am not entirely sure this is correct, because there are loaded tasks in memory for this queue, just updating the shardId will be broken..I think we can just shut it down and start a new one

config/config.go
zklgame: I could not get it. Which same value are you talking about?

engine/interfaces.go
zklgame: It's used to override the shardId so we can combile the operation of destroy + create shards into updateShardId.

config/config.go
zklgame: How are you going to implement the network router?

engine/timer_task_queue.go
zklgame: I think the effect is the same. Let's sync offline.

config/config.go
zklgame: I think current desp is clear to tell this is with HTTP config.

config/config.go
```diff
@@ -43,6 +44,10 @@ type (
 		HttpServer HttpServerConfig `yaml:"httpServer"`
 		// Rpc is the config for rpc calls
 		Rpc RpcConfig `yaml:"rpc"`
+		// AsyncAddresses are the addresses for API service to call AsyncServices' internal APIs
+		AsyncAddresses []string `yaml:"asyncAddresses"`
```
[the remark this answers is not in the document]
zklgame: done

config/config.go
zklgame: Moved to database config.

config/config.go
zklgame: done

service/async/service_impl.go
zklgame: Added comment about the key and value to add more context.

engine/interfaces.go
zklgame: Deleted this API since we don't do the updation.

config/config.go
zklgame: Done.

engine/timer_task_queue.go
zklgame: Done.

service/async/interfaces.go
zklgame: Renamed to GetAdvertiserAddressForShard

engine/immediate_task_queue.go
zklgame: Tried in the last two commits, adding close() cause bugs.

longquanzheng (approved):

PR #125, opened 2024-04-02
Add api server into membership

zklgame:
This MR did:
- add api server into membership
- make membershipConfig an independent config

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -29,6 +29,9 @@ type (
 
 		// AsyncService is config for async service
 		AsyncService *AsyncServiceConfig `yaml:"asyncService"`
+
+		// Membership is the config for cluster members
+		Membership *MembershipConfig `yaml:"membership"`
 	}
 
 	DatabaseConfig struct {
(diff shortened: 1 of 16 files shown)


service/api/service_impl.go
```diff
@@ -36,11 +37,14 @@ func NewServiceImpl(
 	visibilityStore persistence.VisibilityStore,
 	logger log.Logger,
 ) Service {
+	membershipImpl := async.NewMembershipImpl(cfg, logger, nil, async.ServerTypeApi)
```
zklgame, line 40: have to put `NewMembershipImpl` inside async package to avoid circular imports.

config/config.go
```diff
@@ -29,6 +28,9 @@ type (
 
 		// AsyncService is config for async service
 		AsyncService *AsyncServiceConfig `yaml:"asyncService"`
+
+		// Membership is the config for cluster members
+		Membership *MembershipConfig `yaml:"membership"`
```
zklgame, line 33: Now MembershipConfig is in the same level as ApiServiceConfig and AsyncServiceConfig.

longquanzheng: Nice!

longquanzheng (approved):

config/config.go
```diff
@@ -46,8 +49,9 @@ type (
 		HttpServer HttpServerConfig `yaml:"httpServer"`
 		// Rpc is the config for rpc calls
 		Rpc RpcConfig `yaml:"rpc"`
-		// AsyncServiceAddresses are the addresses for API service to call AsyncServices' internal APIs
-		AsyncServiceAddresses []string `yaml:"asyncServiceAddresses"`
+		// AsyncServiceAddress is the address for API service to call the AsyncService's internal APIs
+		// It's required in the standalone mode, but not needed in the cluster mode
```
longquanzheng, line 53: This makes me feel like we probably can deprecate standalone mode, or in other words, it can become automatic somehow in a smart way

zklgame: It's doable, it means the standalone mode will be equal to the cluster mode with only 1 async server, and we need to use memberlist anyway.

PR #511, opened 2026-09-20
Run and Inbox modes for Dex Web v2

longquanzheng:
Two modes for Dex Web v2, after a process has been developed: a semantic **Run** view for an admin, and an **Inbox** for a participant who has to act. Chat is not part of this — it was a mock in the prototype and Dex has not enabled it.

## What the branch does

**Run view — the primary admin view.** A canvas of the flow with a right-hand drawer per run. Non-technical by default: each Step says what it does in the flow's own terms, sourced from `dex:explanation` rather than from prose kept beside the code. Selecting a run with an open Action focuses the Step it is stopped at and opens the drawer; closing resets. Clicking a Step with no run selected still explains the Step, because a definition is a readable artifact on its own.

**Deep Dive — the v1 detailed view, nested rather than replaced.** Its layout and information density are kept intact; only colour and typography are unified, in a `.v2-debug` scoped layer. Reached from the run header, so v1's technical depth is one click away instead of gone.

**Inbox — the participant list.** Shares one `RunList` with the Run view so the two cannot drift into two visual languages. The role contract is not in this branch: it still lists runs rather than "what needs *me*".

**Two human gates in the refund example.** `RequestHumanApprovalStep` and a new `ConfirmCustomerMessageStep`, so multiple actionable Steps per run is exercised by a real flow rather than asserted. `customer-email` and `refund-amount` are now Indexed Attributes; amount is in dollars.

**Canvas layout is governed by measured principles.** `layoutQuality.ts` declares what a drawing must achieve, each with a metric and a limit, and the corpus test runs every shipped definition in all four view modes against them. This replaced tuning constants against one screenshot at a time.

## Review feedback addressed

**1. Restore and simplify run search.** Status and time are the default view — the two questions every Flow answers whether or not it declared anything indexable. An exact run id and one attribute comparison sit behind `Advanced search`. The operator list is derived from the index type, so the UI cannot compose a query the server rejects: `contains` only for fulltext, ordering only for int, double and datetime. The value control follows the Attribute's value type. No keyword search over payloads, inputs, outputs or history — only declared Indexed Attributes, named in the picker from the contract. No multi-row query builder.

Checked against what `FilterBuilder` had before it was removed: operators, the bool select and the datetime input are restored, and **run-id lookup is restored** — it had been lost, and searching for a known run is among the most common reasons to open the panel. `in` stays out; it needs a comma convention in placeholder text and a second value control as soon as anybody wants two ranges.

**2. Multiple human-actionable Steps per run.** Derived from the graph, not declared: a Step is actionable when something outside the flow publishes to the Channel it waits on. A flow that grows a third gate grows a third entry with no UI change. The drawer shows a progress strip, defaults to the first, marks the current one, and **advances to the next pending gate on completion** instead of closing — closing was right when one Action ended the run, but with another gate pending the reader should not have to find it again. Nothing requires navigating back into run history.

**3. Existing Dex interaction model preserved.** The Run view, Process Canvas and Inbox are not redesigned. Changes are additive or fixes to what was already there.

## Fixes found along the way

- **Edges were invisible.** The `stroke-width` rule lost to `@xyflow/react`'s own stylesheet at equal specificity, so every edge sat at the 1px library default — the earlier 1.5px had always been a no-op. Auto-focusing a Step also reused the reader's selection path and dimmed 43 of 47 edges.
- **Region merging under-converged.** The greedy merge bounded its passes by the live `rects.length`, which shrinks on every merge while the counter rises. A group spanning seven rows stopped after four merges and shipped three regions where one was available. This affected every flow, not just the one on screen.
- **Lanes are assigned per group, not per Step.** Reachability classifies a Step correctly and still drew Failure as five regions in three columns, because only three of its members are reached by failing. A group that is mostly error handling is error handling, so all of it shares the recovery lane.
- **Changing Indexed Attributes on a store that already holds runs of that Flow type stops the Worker before it binds its port** — the symptom is silence, not an error. Recorded in `examples/go/README.md`.
- Dead surface removed: `describeFilters`, `QUEUE_COPY.scope`, `unfilteredHint` and `onThisPage` were reachable only from one test, which is why nothing flagged them.

## Verification

- `web`: 208 tests, `npm run check` clean.
- `docs`: `npm run check` clean — both locales build, 156 indexable routes, 7 redirects, SEO audit passed.
- `make copyright-check` clean.
- All four operators, run-id lookup and both gates exercised against a live server: `amount >= 99999` → 0 rows, `>= 0` → 5, `email contains northwind` → 2, `contains nosuchdomain` → 0, `runId = adv-1` → 1.
- Layout guardrails were each **broken on purpose** to confirm they fail. Three earlier attempts at one were vacuous, so a green guardrail is not taken as evidence.

## Known gaps, deliberately not in this branch

- **The role contract.** `dex:field role:` and `dex:role-attribute`, a role picker, and "needs my action" ordering. Without it Inbox is a run list, not a participant view.
- **Region principles are thinly covered.** Only 2 of 33 shipped definitions declare groups — `dex:group` is FDG 2.0 and the other 31 are 1.0 — so `tight-regions` and `group-cohesion` are exercised by 8 of 132 scenes. Pinned by name so stale regenerated JSON cannot quietly empty them. Porting a second FDG 2.0 flow is the way to widen it.
- Multi-role validation needs the 18-step, 5-party onboarding flow ported to Go.
- `dex-web.mdx` and its zh-Hans mirror are not updated for the two modes.


zzheng-aiplay:
Reopened with the review feedback addressed. Since the close:

- **Run search, restored and simplified.** Status and time by default; exact run id and one attribute comparison behind `Advanced search`. Operators come from the index type, so the UI cannot compose a query the server rejects — `contains` only for fulltext, ordering only for int, double and datetime. Only declared Indexed Attributes, never payloads, inputs, outputs or history. No multi-row builder. Run-id lookup is back; the earlier version had lost it.
- **Multiple human-actionable Steps per run.** The refund example gained a confirm-or-rewrite gate for the customer message, so this is exercised by a real flow. Actionable Steps are derived from the graph rather than declared, and the drawer advances to the next pending gate on completion instead of closing.
- **Existing interaction model preserved.** Run view, Process Canvas and Inbox are not redesigned.

Also fixed along the way: Canvas edges were invisible because the `stroke-width` rule lost to `@xyflow/react`'s stylesheet at equal specificity, and group region merging bounded its passes by a shrinking length, so it silently stopped after about half the merges available in any flow.

CI is green (33 pass, 1 skipping). The description above is rewritten in full, including the gaps I deliberately left out — the role contract is not in this branch, so Inbox is still a run list rather than a participant view.

